PR #158, opened 2025-09-04
Docker support

chrisguida:
Adds support for docker

Todo:

- [x] Dockerfile for client + add client to docker-compose
- [x] Ability to set phoenixd-url programmatically rather than hardcode
- [x] Precompile client so it's ready for use rather than compiling on-the-fly
- [x] Add `make run` and `make run-rebuild` to install all the dependencies for docker-compose and optionally build all the images, then run them with `docker-compose up` or `docker-compose up --build`
- [x] Add documentation for `make run` 

For later:
- [ ] Multiarchitecture docker image for optimal image size
- [ ] Cache server image gradle build so we don't recompile from scratch every time

### DIFF
--- a/server/app/src/main/kotlin/pos/ambrosia/services/PhoenixService.kt
+++ b/server/app/src/main/kotlin/pos/ambrosia/services/PhoenixService.kt
@@ -15,9 +15,13 @@ import kotlinx.serialization.json.Json
 import pos.ambrosia.config.AppConfig
 import pos.ambrosia.models.Phoenix.*
 import pos.ambrosia.utils.*
+import io.ktor.server.application.ApplicationEnvironment
 
 /** Service for interacting with Phoenix Lightning node */
-class PhoenixService(private val phoenixUrl: String = "http://phoenixd:9740") {
+class PhoenixService(app: ApplicationEnvironment) {
+  private val config = app.config
+  private val phoenixdUrl = config.property("phoenixd-url").getString()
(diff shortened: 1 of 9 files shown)


server/app/src/main/kotlin/pos/ambrosia/services/PhoenixService.kt
JordyPirata, line 1: For the implementation of flags, we are using [Clickt](https://ajalt.github.io/clikt/options/), which is a library that helps pass variables from the console to the `phoenixdService`.

I recommend configuring the app the same way we pass the secret in [Ambrosia.kt](https://github.com/btcgdl/Ambrosia-POS/blob/main/server/app/src/main/kotlin/pos/ambrosia/Ambrosia.kt). For more information, please see the [Ktor Environments](https://ktor.io/docs/server-configuration-file.html#read-configuration-in-code) documentation.

ca-ruz:
Additional note: the docker container wouldn't start because the port was conflicting with itself.

```
$ git diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 81134d3..a6d4607 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -36,7 +36,7 @@ services:
     depends_on:
       - ambrosia
     ports:
-      - "3000:3000"
+      - "3001:3000"
     environment:
         NEXT_PUBLIC_API_URL: http://ambrosia:9154
 volumes:
```

PR #174, opened 2025-09-19
make client image smaller

chrisguida:
This shrinks the client image from 2.36GB to 1.2GB

The client build is now multistage to minimize runtime image size.

We move scripts/package-client.sh to client/package-client.sh (to be able to use in the docker build) and add a `NO_ZIP` env var to tell the script to just leave the dist folder in /tmp without deleting it, so we can copy it from the build stage into the run stage.


JordyPirata:
Tested :+1:

PR #176, opened 2025-09-19
Fix running server within tests

ca-ruz:
Fix running tests without the need to run the server in the background first


JordyPirata:
tested :+1:

PR #212, opened 2025-10-16
server: add seed endpoint

chrisguida:


JordyPirata:
Super Chris thanks!! :+1:

PR #244, opened 2025-11-14
Feature/initial setup

Sharmaz:
Changes:
- Added PIN Input
- Added Password and PIN tooltip info
- Fixed RFC regex


Vidarte-Alberto (approved):
Sweet!

PR #255, opened 2025-11-24
Feature/store products

Sharmaz:
Changes:
- Add store products view
- Add store products table
- Add store products add product modal
- Add store products edit product modal
- Add store products delete product modal

### DIFF
--- a/client/src/lib/modules.js
+++ b/client/src/lib/modules.js
@@ -220,7 +220,7 @@ export const modules = {
         requiresAuth: true,
         requiresAdmin: false,
         types: ["store"],
-        default: true,
+        default: false,
       },
       {
         path: "/store/products",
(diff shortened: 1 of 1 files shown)


client/src/lib/modules.js
```diff
@@ -222,6 +222,14 @@ export const modules = {
         types: ["store"],
         default: true,
       },
+      {
+        path: "/store/products",
+        component: "Products",
+        requiresAuth: true,
+        requiresAdmin: false,
+        types: ["store"],
+        default: true,
```
Vidarte-Alberto: El default es solo para la pagina que se debe de mostrar en home, entonces en este caso solo deberia de quedarse en default el /store no /stroe/{slug}

PR #269, opened 2025-12-02
Add dynamic secure cookie helper for auth endpoints

sip-21:
Context: Fix cookie inconsistency between login (`secure = false`) and refresh (`secure = true`) endpoints

### Fixed

- **Dynamic Secure Cookies** (server-side): Implemented `shouldUseSecureCookies()` helper in `Authorize.kt`:
  - Automatically uses `secure = false` for HTTP (development/testing)
  - Automatically uses `secure = true` for HTTPS (production)
  - Fixes cookie inconsistency between login (`secure = false`) and refresh (`secure = true`) endpoints
  - Applied to login and refresh endpoints


sip-21:
Resolved in the meantime in this commit:
https://github.com/olympus-btc/ambrosia/pull/282/commits/d9bcd338f0b0afc21cad18b397d46779f651ec8f

PR #273, opened 2025-12-03
add a function in products hook

Vidarte-Alberto:
This pull request adds support for deleting products in the store management interface. The main changes include implementing the `deleteProduct` function in the products hook, updating the products component to use this function, and wiring up the delete confirmation modal.

**Product deletion feature:**

* Added a new `deleteProduct` function to the `useProducts` hook, which sends a DELETE request to the API and refreshes the product list.
* Exposed the `deleteProduct` function from the `useProducts` hook so it can be used by components.
* Updated the `Products` component to import and use the new `deleteProduct` function.
* Integrated product deletion into the confirmation modal: when confirmed, the selected product is deleted using the new function.

### DIFF
--- a/client/src/components/pages/Store/Products/Products.jsx
+++ b/client/src/components/pages/Store/Products/Products.jsx
@@ -135,7 +135,6 @@ export function Products() {
           setDeleteProductsShowModal={setDeleteProductsShowModal}
           onConfirm={() => {
             setDeleteProductsShowModal(false);
-            console.log(productToDelete)
             deleteProduct(productToDelete)
           }}
         />


client/src/components/pages/Store/Products/Products.jsx
```diff
@@ -135,6 +135,8 @@ export function Products() {
           setDeleteProductsShowModal={setDeleteProductsShowModal}
           onConfirm={() => {
             setDeleteProductsShowModal(false);
+            console.log(productToDelete)
```
Sharmaz: Remove console logs.

Sharmaz (approved):
LGTM!

PR #274, opened 2025-12-04
Use multi-stage build for self-contained server image

sip-21:
This Dockerfile allows to:
* Build the JAR inside Docker (no JDK needed on host)
* Work on a fresh clone (no pre-built JAR required)
* Only needs Docker installed

Once merged you should be able to just run:

```bash
git clone https://github.com/olympus-btc/ambrosia
cd ambrosia
docker-compose up -d
```
then 

```bash
open http://localhost:3000
```

Build time: ~6-7 minutes (downloads JDK, Gradle, dependencies)


Sharmaz (approved):
LGTM!!

PR #277, opened 2025-12-05
fix client dockerfile for windows.

Sharmaz:


JordyPirata (approved):

chrisguida:
Btw, this is not needed if the user uses WSL, because that will create the correct Linux line endings. This is only for users who do the `git clone` _outside_ WSL, such as in windows powershell, where Windows line endings are created.

PR #279, opened 2025-12-06
Fix/onboarding

Sharmaz:
Changes:
- EN default langage
- Step 1 Disable restaurant on Select Bussiness
- Step 2 Change password tooltip description to ("To access the wallet.")
- Step 3 
  - Tax ID not required, only Business Name required
  - USD default currency
  - Currency select fixed


Vidarte-Alberto (approved):
LGTM

PR #281, opened 2025-12-07
fix SSL Problem with phoenix.conf and phoenixd

Vidarte-Alberto:
This pull request updates the default webhook URL configuration in the `Ambrosia` command. The change switches the protocol from HTTPS to HTTP and updates the port number from 9443 to 9154.

Configuration update:

* Changed the default webhook URL in the `Ambrosia` command to use HTTP protocol and port 9154 instead of HTTPS and port 9443 (`Ambrosia.kt`).


Sharmaz (approved):
LGTM!!

PR #298, opened 2025-12-17
issue fixed

Vidarte-Alberto:
This pull request makes a small but important update to how product images are displayed in the `ProductsTable` component. The main change is that product image URLs are now processed through the `storedAssetUrl` utility function before being rendered, which likely improves handling of asset URLs (e.g., adding a CDN prefix or fallback logic).

- Updated the `ProductsTable` component to use the `storedAssetUrl` utility for product image URLs, ensuring consistent and possibly safer asset loading (`client/src/components/pages/Store/Products/ProductsTable.jsx`).
- Imported the `storedAssetUrl` utility into the `ProductsTable` component to support the above change (`client/src/components/pages/Store/Products/ProductsTable.jsx`).


Sharmaz (approved):
It Works!!

PR #327, opened 2026-01-08
Fix/client design consistency

Sharmaz:
## Changes
Fixed design consistency on several components of store, that means that we were implementing different styles for buttons, cards, shadows, borders, inputs.

Sections:
- Settings
- Products
- Orders
- Wallet
- Cart


Vidarte-Alberto (approved):
LGTM!!

PR #347, opened 2026-01-21
fix script postinst with the updated paths.

Sharmaz:
Changes:
- fix script post install for Linux dist files.


JordyPirata (approved):
LGTM!!

PR #356, opened 2026-01-22
add re-sign after pack script for macos arm64 dist file.

Sharmaz:
Changes:
- Added script for sign MacOS ARM binaries.


JordyPirata (approved):
LGTM!!

PR #365, opened 2026-01-23
Fix/iu low height screens

Sharmaz:
Changes:
- Less than 800px height screen resolutions fixes at UI for Onboarding and Store Products and Settings.
- Add currency validation at onboarding.
- Disable restaurant selection.


Vidarte-Alberto (approved):
LGTM!!

PR #373, opened 2026-01-24
clear cart when user logout or unauthorized

Vidarte-Alberto:
This pull request updates the authentication logic to ensure that the user's cart data is cleared from local storage when they log out or when their session expires. This helps maintain user privacy and prevents stale cart data from persisting across sessions.

**Authentication and Cart Management Improvements:**

* Added a `clearCartStorage` function to remove the cart from local storage using the `CART_STORAGE_KEY` constant. This function is called during logout and when the session expires to ensure cart data is cleared.
* Updated the `logout` function to call `clearCartStorage` before resetting the authentication state.
* Updated the session expiration handler (`handleExpired`) to call `clearCartStorage` when the session expires.

**Imports and Dependencies:**

* Imported `CART_STORAGE_KEY` from the cart hook to support cart data removal.

### DIFF
--- a/client/src/modules/auth/AuthProvider.jsx
+++ b/client/src/modules/auth/AuthProvider.jsx
@@ -1,8 +1,9 @@
-﻿"use client";
+"use client";
 import { createContext, useEffect, useState } from "react";
 import { loginFromService, logoutFromService } from "./authService";
 import { apiClient } from "../../services/apiClient";
 import { useRouter } from "next/navigation";
+import { CART_STORAGE_KEY } from "../../components/pages/Store/Cart/hooks/usePersistentCart";
 
 export const AuthContext = createContext();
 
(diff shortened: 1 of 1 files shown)


client/src/modules/auth/AuthProvider.jsx
```diff
@@ -1,8 +1,9 @@
-﻿"use client";
+"use client";
 import { createContext, useEffect, useState } from "react";
 import { loginFromService, logoutFromService } from "./authService";
 import { apiClient } from "../../services/apiClient";
 import { useRouter } from "next/navigation";
+import { CART_STORAGE_KEY } from "../../components/pages/Store/Cart/hooks/usePersistentCart";
```
Sharmaz: Add `@` imports

client/src/modules/auth/AuthProvider.jsx
```diff
@@ -1,8 +1,9 @@
-﻿"use client";
+"use client";
 import { createContext, useEffect, useState } from "react";
 import { loginFromService, logoutFromService } from "./authService";
 import { apiClient } from "../../services/apiClient";
```
Sharmaz: Add `@` imports

Sharmaz (changes requested):
Don't forget the imports

Sharmaz (approved):

PR #375, opened 2026-01-27
remove appDmg dependency.

Sharmaz:
Changes:
- Fix create dist files from Linux SO
- Fix afterPack permissions


Vidarte-Alberto (approved):
LGTM!!!

PR #384, opened 2026-01-28
remove back button for first step in onboarding

Vidarte-Alberto:
This pull request updates the onboarding flow UI and its related tests to improve button rendering logic and alignment. The main changes involve adjusting when the "Back" button is displayed and updating the test to match this new behavior.

UI improvements in onboarding flow:

* The "Back" button in the `Onboarding` component is now only rendered when `needsBusinessType` is false and the current step is not the first step (`step !== 1`). Previously, it was always rendered and just disabled on the first step. This change improves clarity and reduces unnecessary UI elements.
* The layout of the button container in the onboarding footer has been updated to use `w-full` for full width and `ml-auto` to align the "Next" button to the right, improving visual alignment.
* Added a closing `</div>` to properly wrap the component's content, ensuring correct structure and rendering.

Test updates:

* The onboarding test now checks that the "Back" button is not present in the DOM on the first step, instead of checking that it is disabled. This aligns the test with the new button rendering logic.

### DIFF
--- a/client/src/components/pages/Onboarding/Onboarding.jsx
+++ b/client/src/components/pages/Onboarding/Onboarding.jsx
@@ -185,49 +185,50 @@ export function Onboarding() {
 
           <Divider className="my-8 bg-gray-400" />
 
-          <div className="flex justify-between">
-            {!needsBusinessType && (
+          <div className="flex w-full">
+            {(!needsBusinessType && step !== 1) && (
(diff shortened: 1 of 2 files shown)


Sharmaz (commented):
This is not a hotfix, the name of the branch is not descriptive, the issue number or link should be at PR description.

PR #389, opened 2026-01-31
fix(onboarding): remove back button from onboarding first step

### LINKED ISSUE #383: Back button in stepper
On the first step of the stepper flow, the Back button is currently visible.
Since there is no previous step to navigate to, this button should not be displayed at all on step 1.
Expected behavior:
The Back button should be hidden on the first step of the stepper.
Actual behavior:
The Back button is visible on step 1.
This creates unnecessary UI noise and may confuse users by suggesting a navigation option that does not exist. 

Image

Vidarte-Alberto:
This pull request updates the onboarding flow's navigation buttons to improve their behavior and fixes related tests. The most notable changes are adjustments to when the "Back" button is shown and enabled, as well as corresponding updates to the test suite.

**Onboarding Button Logic Improvements:**

* The "Back" button in `Onboarding.jsx` is now only rendered when the user is not on the first step and does not need to select a business type, instead of being rendered and disabled on the first step. This provides a cleaner UI and avoids showing a disabled button unnecessarily.
* The button group layout was adjusted to ensure proper alignment, with the "Next" button always aligned to the right.

**Test Updates:**

* The test for the "Back" button was updated to check that the button is not present in the DOM on the first step, instead of checking if it is disabled. This matches the new UI logic.

**Other:**

* Minor JSX structure fix: closed a `div` element to ensure valid markup.

### DIFF
--- a/client/src/components/pages/Onboarding/__tests__/Onboarding.test.jsx
+++ b/client/src/components/pages/Onboarding/__tests__/Onboarding.test.jsx
@@ -86,7 +86,7 @@ describe("Onboarding Wizard", () => {
     await act(async () => {
       renderOnboarding();
     });
-    expect(screen.getByText("buttons.back")).toBeDisabled();
+    expect(screen.queryByText("buttons.back")).not.toBeInTheDocument();
   });
 
   it("disables the Next button if Pin not added", async () => {
--- a/client/src/components/pages/Onboarding/Onboarding.jsx
+++ b/client/src/components/pages/Onboarding/Onboarding.jsx
@@ -185,49 +185,50 @@ export function Onboarding() {
 
           <Divider className="my-8 bg-gray-400" />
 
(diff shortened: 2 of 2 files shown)


client/src/components/pages/Onboarding/__tests__/Onboarding.test.jsx
```diff
@@ -86,7 +86,7 @@ describe("Onboarding Wizard", () => {
     await act(async () => {
       renderOnboarding();
     });
-    expect(screen.getByText("buttons.back")).toBeDisabled();
+    expect(screen.queryByText("buttons.back")).not.toBeInTheDocument();
```
Sharmaz, line 89: Why `queryByText` instead of `getByText` ?

Sharmaz (approved):
LGTM!!

PR #394, opened 2026-02-15
Add regression tests for payment infrastructure and E2E flows

themikemoniker:
## Summary
- Add 102 client unit tests covering `apiClient` (token refresh, error handling), `bitcoinPriceService` (caching, fiat/sats conversions), `usePaymentWebsocket` (WS lifecycle, reconnection, payment events), and `modules` (routing, permissions, business type filtering)
- Add 29 Python E2E tests for products CRUD, orders lifecycle, payments/webhook HMAC validation, and multi-user permission boundaries
- Add 14 Playwright browser tests for login, products, cart-checkout (NWC-critical), and wallet pages

## Test Results

| Suite | Passed | Skipped | Failed |
|-------|--------|---------|--------|
| Client unit tests (Jest) | 670 | 0 | 0 |
| Python E2E (pytest) | 46 | 4 | 0 |
| Playwright browser tests | 14 | 0 | 0 |
| **Total** | **730** | **4** | **0** |

### Key fixes to align tests with actual Ktor API schemas

1. **`/users/me` response parsing** — response is `{"user": {"user_id": "..."}, "perms": [...]}`, not a flat `{"id": "..."}`
2. **Order status values** — must be `"open"`, `"closed"`, or `"paid"` (not `"pending"`)
3. **Dishes vs Products** — order dishes reference the `dishes` table, not `products`; creating a dish requires `POST /dishes` with `{name, price, category_id}`
4. **conftest.py** — handle `"already exists"` error message variant from the server (in addition to `"UNIQUE constraint failed: users.name"`)
5. **Product payload** — include all required fields: `SKU`, `category_id`, `quantity`, `min_stock_threshold`, `max_stock_threshold`, `cost_cents`, `price_cents`
6. **Payment endpoints** — use `/payments/methods`, `/payments/currencies`, and `/payments/ticket-payments` (not `/payment-methods`)

### Playwright fixes

- Added `global-setup.js` that initializes the Ktor database via `POST /initial-setup` before browser tests run (uses unique username to avoid collision with existing E2E test users)
- Fixed auth fixture and login tests to match actual PinLogin UI (user dropdown + number pad, not text input)
- The Next.js middleware redirects to `/onboarding` when the DB config is missing — global setup ensures config exists

### 4 skipped Python E2E tests (server-side limitations)

- **Payment creation** (2 tests) — returns 500 due to DB connection issue in test environment
- **Payment-to-ticket linking** (1 test) — depends on payment creation
[...]

PR #397, opened 2026-02-16
Fix e2e lint errors in regression tests

themikemoniker:
## Summary
- Remove `.js` extensions from relative imports in e2e specs (`import/extensions` rule)
- Use double quotes for selector strings in `login.spec.js` (`quotes` rule)
- Suppress false-positive `react-hooks/rules-of-hooks` on Playwright's `use()` fixture function in `auth.js`

Fixes the 6 ESLint errors that cause CI to fail on PR #394.

## Test plan
- [x] `npm run lint` passes locally (0 errors, 5 warnings)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/client/e2e/cart-checkout.spec.js
+++ b/client/e2e/cart-checkout.spec.js
@@ -1,4 +1,4 @@
-import { test, expect } from "./fixtures/auth.js";
+import { test, expect } from "./fixtures/auth";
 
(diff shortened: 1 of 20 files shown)


themikemoniker:
not sure if this is needed, because other branches are passing the cicd  without this

PR #401, opened 2026-02-18
Update workflow commands for ktlint main and test checks

ca-ruz:
Updated github workflow to run checks for both main and test files.


JordyPirata (approved):
LGTM!!

PR #414, opened 2026-02-25
fix auth loop caused by auth flow.

Sharmaz:
## Description
Fix infinite redirect loop on Electron startup caused by stale authentication cookies. When a previous session's refreshToken cookie was present but expired, the middleware and page-level redirect logic created a cycle that prevented users from reaching the login screen.

## Changes:
  - Remove middleware redirect from /auth → / based on cookie presence alone, breaking the stale-cookie loop
  - Add client-side redirect in PinLogin when user is already authenticated, replacing the removed middleware behavior
  - Change post-login navigation from window.location.reload() to window.location.replace("/") to avoid reloading on /auth and flashing the login screen
  - Fix httpClient to not dispatch auth:expired event when skipRefresh: true is set, preventing silent requests from triggering session expiry handlers
  - Fix AuthProvider.fetchUser to use skipRefresh: true on initial mount, preventing a refresh cycle on unauthenticated load
  - Fix useTurn to gate API calls behind isAuth, preventing unauthenticated requests on mount
  - Fix authService.getUsers to properly forward silentAuth as skipRefresh instead of ignoring the parameter
  - Simplify LoadingCard by removing unused message prop — component always uses its own translation
  - Simplify page.jsx to render <LoadingCard /> without hardcoded message strings that were never displayed


Vidarte-Alberto (approved):
LGTM!

PR #416, opened 2026-02-26
Chore/electron improvements

Sharmaz:
## Changes:
- Fix splash window security model: replace nodeIntegration/contextIsolation:false with a dedicated splash-preload.js using contextBridge, matching the main window's sandbox model
- Remove secrets from CLI args in PhoenixdService and BackendService to prevent exposure via `ps aux`; pass sensitive values as environment variables instead
- Add SHA256 integrity verification for all downloaded binaries (phoenixd, JRE, Node.js) before extraction using a shared verify-checksum.js utility; fetches checksums from official sources (ACINQ SHA256SUMS.asc, Adoptium assets API, nodejs.org SHASUMS256.txt)
- Replace hardcoded JAR filename in resourcePaths.js with dynamic detection using fs.readdirSync, preventing build failures on version bumps
- Write config files (ambrosia.conf, phoenix.conf) with mode 0o600 to prevent world-readable secrets on multi-user systems
- Add background@2x.png for macOS DMG installer Retina display support and resize background.png to correct 1x logical pixel dimensions (660×450)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #421, opened 2026-02-27
feat(backend): simplify webhook logic with docker flag and updated envs

JordyPirata:
- Add --docker flag and IS_DOCKER env var to handle internal networking.
- Update Dockerfile to include --docker flag by default.
- Fix invalid webhook URLs by defaulting to 127.0.0.1 when binding to 0.0.0.0/::.
- Rename webhook host override to PHOENIXD_WEBHOOK_URL for full control.
- Inline webhook URL generation logic and remove redundant function.
- Add missing envvar mappings for secret, password, and webhook-secret options.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -29,4 +29,4 @@ COPY --from=builder /build/app/build/libs/ambrosia-*.jar ./ambrosia.jar
 
 EXPOSE 9154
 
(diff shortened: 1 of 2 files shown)


Sharmaz (approved):
LGTM!!

PR #430, opened 2026-03-09
Standardization of API Responses and Business Logic

ca-ruz:
1. Critical Bug Fixes (HTTP 500)
   * Serialization Fix: Resolved a 500 Internal Server Error in Orders.kt (/total-sales) where the Kotlin JSON serializer crashed when handling a generic Map with mixed types (String and Double). Fixed by converting numeric values to Strings before responding.
   * Global Exception Handling: Updated Handler.kt to reorder exception priorities, ensuring specific custom exceptions are caught before generic Throwable or SQLException handlers.

  2. Enhanced Business Logic ("Missing" vs. "Empty")
   * Service Layer Updates: Modified OrderService, TableService, CategoryService, PermissionsService, and TicketPaymentService to return null instead of emptyList() when a parent entity (User, Table, Role, Space, etc.) is missing from the database.
   * API Logic: Updated filter endpoints across the api/ folder to distinguish between:
       * 404 Not Found: The parent entity provided in the URL does not exist.
       * 200 OK: The parent entity exists, but has no associated records (returns an empty list with a descriptive message).

  3. HTTP Status Code Standardization
   * DELETE Standard: Reverted all DELETE endpoints from 200 OK to a strict 204 No Content standard (removing response bodies) to align with RESTful best practices.
   * GET Standard: Switched "empty result" responses from 204 No Content to 200 OK so that custom informative messages (e.g., "No orders found") are visible in the response body for easier debugging and frontend handling.
   * PUT/POST Consistency:
       * Updated Currency.kt and InitialSetup.kt to return 404 (instead of 400) for unknown currency acronyms.
       * Standardized PUT failures to return 404 when the target ID is missing.
   * Webhook Ack: Changed Phoenix webhook acknowledgment from 202 Accepted to 200 OK ("Ok") to match its synchronous processing nature.

  4. Test Suite Improvements
   * Fixed Failing Tests: Updated CategoryServiceTest, OrderServiceTest, TableServiceTest, and TicketPaymentServiceTest to handle the new nullable return types from services.
   * Null-Safety in Tests: Added assertNotNull checks to "Happy Path" tests to ensure compiler safety.
   * New Test Cases: Added negative test cases (e.g., returns null when user not found) to validate the new "Missing vs. Empty" logic.

  5. Repository Management
   * Conflict Resolution: Manually resolved complex merge conflicts in Handler.kt (Exception mappings) and Shifts.kt (Permissions vs. Logic changes).

### DIFF
--- a/server/app/src/main/kotlin/pos/ambrosia/api/Authorize.kt
+++ b/server/app/src/main/kotlin/pos/ambrosia/api/Authorize.kt
@@ -97,7 +97,7 @@ fun Route.auth(
         val accessTokenResponse = tokenService.generateAccessToken(userInfo)
         val refreshTokenResponse = tokenService.generateRefreshToken(userInfo)
 
-        val perms = permissionsService.getByRole(userInfo.roleId)
+        val perms = permissionsService.getByRole(userInfo.roleId) ?: emptyList()
         if (perms.isEmpty()) {
             logger.info("The user doesn't have a permissions")
             call.respond(HttpStatusCode.Forbidden)
--- a/server/app/src/main/kotlin/pos/ambrosia/api/Categories.kt
+++ b/server/app/src/main/kotlin/pos/ambrosia/api/Categories.kt
@@ -34,7 +34,7 @@ fun Route.categories(service: CategoryService) {
             }
             val items = service.getCategories(type)
             if (items == null) {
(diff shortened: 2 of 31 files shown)


Sharmaz (approved):
LGTM!!

JordyPirata (approved):

PR #431, opened 2026-03-10
Feature/settings auto liquidity

### LINKED ISSUE #276: [feat]: Add phoenixd autoLiquidity configuration to setup wizard

Sharmaz:
## Changes:
  - Add Auto Liquidity toggle to Settings page (Electron only) — reads and writes `auto-liquidity` in `phoenix.conf` via IPC
  - Add `phoenixd:get-auto-liquidity` and `phoenixd:set-auto-liquidity` IPC handlers in Electron main process
  - Restart phoenixd automatically after config change, skipping restart in dev mode
  - Add restart lock to prevent concurrent phoenixd restarts from rapid UI interaction
  - Add 500ms debounce on the toggle to avoid redundant IPC calls
  - Update switch state optimistically — flips immediately, reverts on failure
  - Show contextual toasts: success on restart, warning when phoenixd is external and requires manual restart, danger on error
  - Fix race condition in `PhoenixdService`: `close` event handler now checks process identity before cleanup to avoid nullifying a newly started process
  - Fix `stop()` to wait for the actual process exit event before resolving, preventing port conflicts between stop and start
  - Fix `killByPort` to use `-sTCP:LISTEN` (macOS/Linux) so only the listening process is killed — prevents accidentally killing Electron itself via its pooled HTTP connections
  - Add `externalServices.phoenixd = false` after a successful restart in `ServiceManager` so subsequent toggles work within the same session
  - Add `isElectron` utility in `@lib/isElectron` for safe SSR-compatible Electron detection
  - Add `lightning` i18n keys in English and Spanish locales
  - Add tests for `isElectron`, `useAutoLiquidity` hook, `LightningCard` component, and `LightningCard` visibility in `Settings`

### DIFF
--- a/client/src/components/pages/Store/Settings/Lightning/LightningCard.jsx
+++ b/client/src/components/pages/Store/Settings/Lightning/LightningCard.jsx
@@ -1,0 +1,62 @@
+"use client";
+
+import { addToast, Card, CardBody, CardHeader, Switch } from "@heroui/react";
+import { useTranslations } from "next-intl";
+
+import { useAutoLiquidity } from "@/hooks/useAutoLiquidity";
+
+export function LightningCard() {
+  const t = useTranslations("lightning");
+  const { enabled, loading, restarting, toggle } = useAutoLiquidity();
+
+  const handleToggle = async (newEnabled) => {
+    const result = await toggle(newEnabled);
+    if (result === "manual") {
+      addToast({ color: "warning", description: t("manualRestartRequired") });
+    } else if (result) {
+      addToast({ color: "success", description: t("restartSuccess") });
+    } else {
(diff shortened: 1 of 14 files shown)


Vidarte-Alberto (approved):
LGTM!

ca-ruz (approved):
LGTM!

Issue #411, opened 2026-02-20
UX Improvements


Sharmaz:
### Description:
- Users don't know/find how/where to sell their products. We need to find a better name/indicator to sell. Replace `Cart` and use another word `Sell` for example.
- Define where / how add the create channel mesage, how to lead the user after the initial setup to create a channel for start selling. `Driverjs` ✅ 
- Lightning Payment method by default
- Cart cash modal does not change the calculation change when user enters a number until unfocus number input

PR #432, opened 2026-03-11
Chore/client UI improvements

Sharmaz:
## Changes:
  - Default BTC Lightning as the selected payment method in the cart selector on load
  - Fix payment method deselection — clicking the same item twice no longer clears the selection
  - Fix CashPaymentModal NumberInput — amount now updates in real-time while typing, not only on blur
  - Rename Cart → Sale / Venta in navbar label and section title (routes and components unchanged)
  - DriverJS tour styles — consistent design system styles via globals.css using Tailwind CSS custom properties, no more per-component imports
  - DriverJS tour refactor — generic useTour hook, tour logic removed from layout and wallet components
  
  ## Issue
  https://github.com/olympus-btc/ambrosia/issues/411

### DIFF
--- a/client/src/app/globals.css
+++ b/client/src/app/globals.css
@@ -1,4 +1,5 @@
 @import "tailwindcss";
+@import "driver.js/dist/driver.css";
 @plugin './hero.ts';
 /* Note: You may need to change the path to fit your project structure */
 @source '../../node_modules/@heroui/theme/dist/**/*.{js,ts,jsx,tsx}';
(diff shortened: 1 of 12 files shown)


ca-ruz (approved):
LGTM !

client/src/app/globals.css
```diff
@@ -1,4 +1,5 @@
 @import "tailwindcss";
+@import "driver.js/dist/driver.css";
 @plugin './hero.ts';
 /* Note: You may need to change the path to fit your project structure */
```
Vidarte-Alberto: Please remove the unnecesary comments

client/src/app/globals.css
```diff
@@ -175,3 +176,90 @@
     );
   }
 }
+
+/* ── DriverJS — Ambrosia design overrides ── */
```
Vidarte-Alberto, line 177: No comments

Vidarte-Alberto (changes requested):

client/src/app/globals.css
Sharmaz: I touch that file your right, however...

client/src/app/globals.css
Sharmaz: ...That is neccesary to know where driverjs styles begin on the global CSS

Vidarte-Alberto (approved):
LGTM!!

PR #435, opened 2026-03-12
change host at electron services and client, add fallback.

Sharmaz:
## Changes:
  - Fixed Lightning payment WebSocket not detecting payments after checkout
  - Replaced hardcoded `127.0.0.1` default with `window.location.hostname` fallback in `getWsUrl()` so the WebSocket always connects to the same host the browser loaded from, ensuring auth cookies are sent correctly
  - Updated `ServiceManager.js` and `NextJsService.js` to use `localhost` instead of `127.0.0.1` when passing the backend host to Next.js in Electron mode
  - Preserved explicit `NEXT_PUBLIC_WS_URL` and `NEXT_PUBLIC_API_URL` env vars as higher-priority overrides for Docker and custom deployments
  - Root cause: commit `fad593d` (Jan 29) replaced `window.location.hostname` with a hardcoded `127.0.0.1` default — browsers don't send cookies set on `localhost` to requests targeting `127.0.0.1`, causing the WebSocket auth to fail silently

### DIFF
--- a/client/src/config/api.js
+++ b/client/src/config/api.js
@@ -1,4 +1,4 @@
-const DEFAULT_API_URL = "http://127.0.0.1:9154";
+const DEFAULT_API_URL = "http://localhost:9154";
 
 export function getApiUrl() {
   return process.env.NEXT_PUBLIC_API_URL || DEFAULT_API_URL;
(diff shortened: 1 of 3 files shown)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #437, opened 2026-03-12
Chore/electron improvements

Sharmaz:
## Changes:
  - Bind Next.js to 127.0.0.1 instead of 0.0.0.0 to prevent unintended LAN exposure
  - Add 2-minute global startup timeout to startAll() with error propagation
  - Add 30-second post-startup health monitor that emits service:error if a service dies
  - Add guard in restartService() to skip external services and avoid null-process errors
  - Add redirect loop protection (max 5 redirects) in download-phoenixd.js and download-jre.js
  - Remove process listeners on cleanup() in BackendService and PhoenixdService to prevent listener accumulation on restart
  - Refactor healthCheck.js into a single checkService() function eliminating triplicated code
  - Add res.resume() in all health check HTTP responses to prevent stream leaks
  - Replace BIP39 mnemonic with crypto.randomBytes(32) for JWT and webhook secrets
  - Whitelist environment variables passed to Next.js child process (remove full process.env spread)
  - Add conditional logger (logger.js) that suppresses console.log in production unless DEBUG=true
  - Centralize magic numbers (timeouts, port ranges, versions) into constants.js
  - Add IPC argument sanitization in preload.js to strip functions, symbols and circular references
  - Add assertExists() validation in resourcePaths.js to fail fast with clear errors on missing binaries
  - Add parallel JRE downloads on Windows ARM64 builds using Promise.all()
  - Add visual error state in splash screen (splash:error event) showing failed service and error message
  - Add 7-day update expiration reminder for downloaded Windows updates in AutoUpdater
  - Fix splash screen IPC by removing sandbox: true from splash BrowserWindow (incompatible with ipcRenderer in Electron 36) and adding splash-preload.js to electron-builder files
  - Fix dev mode splash progress by emitting service:started for Next.js in ServiceManager dev path
  - Wait for did-finish-load before sending IPC to splash to eliminate renderer-not-ready race condition

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -388,7 +388,12 @@ app.whenReady().then(async () => {
 
     serviceManager.on('service:error', ({ service, error }) => {
       console.error(`[Electron] Service error: ${service}`, error);
-      updateSplash(null, null, `Error starting ${service || 'service'}`);
+      if (splashWindow && !splashWindow.isDestroyed()) {
+        splashWindow.webContents.send('splash:error', {
+          service,
+          message: error ? error.message || String(error) : `Failed to start ${service || 'service'}`,
+        });
+      }
     });
 
     serviceManager.on('all:started', () => {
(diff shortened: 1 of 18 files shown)


ca-ruz (approved):
LGTM!

JordyPirata (approved):
LGTM!!

PR #438, opened 2026-03-13
Fix/shifts

Sharmaz:
## Changes:
  - Allow opening a shift with $0 initial cash, enabling Lightning-only merchants who don't manage a cash drawer
  - Detect a stuck/unclosed shift proactively when the "Open Shift Required" modal appears — if the server finds an open shift, the modal closes automatically and the widget activates without requiring user interaction
  - Handle shift_already_open (HTTP 409) silently by calling refreshTurn() instead of showing an error state, letting the modal resolve itself
  - Remove redundant "Shift already open" stuck-state UI from OpenTurnForm and RequireOpenTurn — the widget is the single place to manage an open shift
  - Add tests for proactive shift detection: verifies checkOpenShift is called on modal open, skipped when loading or shift already loaded, and not called twice on re-render
  - Add tests for $0 validation and silent shift_already_open handling

### DIFF
--- a/client/src/components/turn/OpenTurnForm.jsx
+++ b/client/src/components/turn/OpenTurnForm.jsx
@@ -10,10 +10,10 @@ import { useTranslations } from "next-intl";
 import { useTurn } from "@/hooks/turn/useTurn";
 
 export default function OpenTurnForm({ onOpened }) {
-  const [initialAmount, setInitialAmount] = useState(1);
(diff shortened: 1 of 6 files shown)


ca-ruz (approved):
LGTM !

Vidarte-Alberto (approved):
LGTM!!!

PR #442, opened 2026-03-16
Chore/electron improvements

Sharmaz:
## Changes:
  - Remove bip39 dependency and replace secret generation with crypto.randomBytes(32) in ConfigurationBootstrap
  - Add conditional debug logger (utils/logger.js) and connect it to main.js, all services, healthCheck.js, portAllocator.js, and preload.js — logs are silenced in production unless DEBUG=true
  - Add process.exit(0) to download-jre.js, download-phoenixd.js, and download-node.js to prevent build hang caused by open HTTP connections keeping the Node.js event loop alive
  - Fix Next.js 16 standalone nested directory structure in build-client.js to resolve missing server.js during production build
  - Fix process reference check in BackendService and NextJsService — capture spawnedProcess before attaching event handlers to prevent stale handlers from corrupting active process state during restarts
  - Add removeAllListeners() in NextJsService.cleanup() to prevent listener accumulation across service restarts
  - Add redirect limit (MAX_REDIRECTS = 5) to download-node.js to prevent infinite redirect loops
  - Skip external services in ServiceManager health monitor to avoid emitting false service:error events for services not owned by the app
  - Add port number validation in PhoenixdService.killByPort() before constructing shell command
  - Log dynamic port allocation error before falling back to default ports in portAllocator.js

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -5,18 +5,19 @@ const { app, BrowserWindow, Menu, dialog, shell, ipcMain } = require('electron')
 const AutoUpdater = require('./services/AutoUpdater');
 const { readConfig, writeConfig } = require('./services/ConfigurationBootstrap');
 const ServiceManager = require('./services/ServiceManager');
+const logger = require('./utils/logger');
 const { getPhoenixDataDirectory } = require('./utils/resourcePaths');
 
 // To prevent multiple instances of the application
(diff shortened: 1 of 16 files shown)


ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):
LGTM!!

PR #443, opened 2026-03-16
add specific scripts for build linux dist files.

Sharmaz:
## Changes:  
- Add individual Linux build scripts for single-format packaging (deb, rpm, tar.gz) for both x64 and arm64 architectures, allowing developers and testers to generate only the format they need without waiting for all three.


ca-ruz (approved):
LGTM!

JordyPirata (approved):
LGTM!!

PR #473, opened 2026-03-26
 Fix: enforce minimum 4-digit PIN and required name validation for users     

### LINKED ISSUE #451: Fix user creation with 1 digit pin
**Describe the bug**
A user can be created with a PIN shorter than 4 digits (e.g. 1 digit) through the Add User / Edit User form. However, PinLogin enforces a minimum of 4 digits, making it impossible for that user to ever log in — they are permanently locked out of the system.

**To Reproduce**
 1. Go to `/store/users`
 2. Click **Add User**
 3. Fill in name, role, and enter a PIN with fewer than 4 digits (e.g. `1`)
 4. Click Add — user is created successfully
 5. Go to `/auth` (PinLogin)
 6. Select the newly created user and enter their 1-digit PIN
 7. Click **Log In** — blocked with "The PIN must be at least 4 digits long"

**Expected behavior**
The system should enforce a minimum PIN length of 4 digits at both layers:
  - Client: The Add User / Edit User form should validate PIN length and show an error before submitting, preventing creation of a user with an invalid PIN.
  - Server: POST /users and PUT /users/:id should reject requests where pin.length < 4 with a 400 Bad Request response.

**Screenshots**

Image

Image

Image

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome
 - Version: Ambrosia POS 0.6.0-Beta-Development

**Additional context**
PIN_LENGTH = 4 is defined as a constant in PinLogin.jsx. The same constraint must be enforced in AddUsersModal.jsx, EditUsersModal.jsx (client) and in the users endpoint handler (server). The
  onboarding step that creates the first user (AddUserAccount.jsx) already has maxLength={4} on the PIN input but also lacks minimum length validation.

ca-ruz:
Adds validation on both frontend and backend to enforce that user name and PIN meet minimum requirements:                
  
- POST /users: rejects blank name, blank PIN, or PIN shorter than 4 digits          
- PUT /users/{id}: rejects blank name or PIN shorter than 4 digits (blank PIN = no change)                                      
- Add modal: name and PIN are required, PIN enforces 4-digit minimum                    
- Edit modal: name is required, PIN is optional but enforces 4-digit minimum if 
  provided                                     
- E2E and unit tests covering all new validation rules 

Fixes #451


client/src/components/pages/Store/Users/EditUsersModal.jsx
Sharmaz, line 64: Add validation and errorMessage for edit user name

Sharmaz, line 94: This also should be required

Sharmaz (changes requested):
Add validations for EditUsersModal

client/src/components/pages/Store/Users/EditUsersModal.jsx
ca-ruz: Added isRequired and errorMessage to the name field. Also added backend validation on PUT /users/{id} to reject blank names, and an E2E test covering that case.

ca-ruz: Kept PIN optional in the Edit modal intentionally, making it required would force users to re-enter their PIN every time they edit any field. The backend already rejects PINs shorter than 4 digits when one is provided.

Sharmaz (approved):
LGTM!!

JordyPirata (approved):
LGTM!!

PR #488, opened 2026-04-10
Fix/mobile copy buttons

### LINKED ISSUE #484: Copy Button on Mobile doesn't copy to clipboard
**Describe the bug**
  Copy buttons do not work on mobile devices. Tapping any "Copy" button appears to trigger the action (toast "Copied / Text copied to clipboard" may appear) but nothing is actually copied to the
  clipboard.

  **To Reproduce**
  1. Open the app on a mobile device (iOS or Android)
  2. Generate a Lightning Invoice or navigate to any section with a copy button
  3. Tap the "Copy" button next to the BOLT11 Invoice (or any other copy button)
  4. Attempt to paste — clipboard is empty

  **Expected behavior**
  Tapping a "Copy" button should copy the associated text to the device clipboard, allowing the user to paste it in another app.

  **Screenshots**
  Lightning Invoice modal on iOS showing "Copied / Text copied to clipboard" toast while clipboard remains empty:
 

Image

  **Desktop (please complete the following information):**
  - N/A — issue is mobile-only

  **Smartphone (please complete the following information):**
  - Device: iPhone and Android
  - OS: iOS and Android
  - Browser: Safari, Chrome, and installed PWA
  - Version: All tested versions

  **Additional context**
  - All copy buttons across the app are affected, not just the Lightning Invoice modal.
  - The `navigator.clipboard.writeText()` API requires a secure context (HTTPS) and, on iOS Safari, requires the call to happen synchronously within a user gesture handler. If the copy action is
  triggered asynchronously (e.g., after an `await` or inside a `Promise.then()`), iOS Safari silently fails or throws a `NotAllowedError`.
  - On Android Chrome, clipboard access may also be blocked if the page does not have focus or if permissions are denied.
  - **Root cause to investigate:** Check all `handleCopy` / clipboard functions in the client — likely `navigator.clipboard.writeText()` is called after an `await`, breaking the user gesture chain on
  iOS. May need to use a synchronous fallback (`document.execCommand('copy')`) or restructure the async flow to call `writeText` before any `await`.
  - **Affected layer:** Client only.

Sharmaz:
## Changes:
- Fix `copyToClipboard` to work on mobile over HTTP: use `window.isSecureContext` to gate the Clipboard API, use native `onClick` instead of `onPress`, and implement iOS-specific text selection via `createRange` within the active modal container
- Refactor `copyToClipboard` into its own module (`utils/copyToClipboard.js`), separating it from `utils/formatters.js`

https://github.com/olympus-btc/ambrosia/issues/484


JordyPirata (approved):
LGMT!!

Vidarte-Alberto (approved):
LGTM!!

PR #490, opened 2026-04-12
feat(auth): replace fixed-window rate limiter with Fibonacci backoff

JordyPirata:
## Summary

- Replaces the fixed-window login rate limiter with a **Fibonacci backoff** strategy: the first 5 failed attempts return `401`, and each subsequent failure blocks the IP for an increasing duration (`fib(1)=1min`, `fib(2)=1min`, `fib(3)=2min`, …, up to ~52 days at `fib(25)`).
- The block counter is **cumulative** and only resets on a successful login, making brute-force increasingly expensive over time.
- Adds `Retry-After` header (in seconds) alongside the `retryAfter` field in the `429` body.
- Marks the rate-limit E2E tests as `slow` so they are skipped in CI by default and only run locally with `--run-slow`.

### `server/app/src/main/kotlin/pos/ambrosia/api/Authorize.kt`

- Replaced `Pair<Int, Long>` state with a typed `IpState(failureCount, blockUntil)` data class.
- Precomputed Fibonacci table (`FIB`) drives the block duration; `fibIndex = failureCount - FREE_ATTEMPTS`.
- `isBlocked` now checks `now < blockUntil` (absolute timestamp) instead of a rolling window.
- `recordFailure` is idempotent on the counter — it no longer resets the window on overflow.
- Removed `InvalidCredentialsException` throw path; replaced with an inline `401 Unauthorized` response for consistency.

### `server/e2e_tests_py/tests/test_rate_limit_e2e.py`

- Three `@pytest.mark.slow` tests covering: grace period + shared bucket across sessions, `Retry-After` header/body contract, and counter reset after a successful login.

### `.github/workflows/e2e.yml`

- Removed `--run-slow` flag; slow tests are excluded from CI to keep the pipeline fast.

## Test plan

- [ ] Run unit tests: `cd server && ./gradlew test`
- [ ] Run fast E2E suite (CI-equivalent): `cd server/e2e_tests_py && pytest tests/ -v --tb=short`
- [ ] Run full E2E suite including rate-limit tests: `pytest tests/ -v --tb=short --run-slow`
- [ ] Manually verify that repeated failed logins on `/auth/login` return `401` for the first 5 attempts, then `429` with `Retry-After` header on the 6th.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -70,4 +70,4 @@ jobs:
       run: |
         cd server/e2e_tests_py
         source .venv/bin/activate
-        pytest tests/ -v --tb=short --run-slow  # Run all tests including slow ones in CI
+        pytest tests/ -v --tb=short
--- a/server/app/src/main/kotlin/pos/ambrosia/api/Authorize.kt
+++ b/server/app/src/main/kotlin/pos/ambrosia/api/Authorize.kt
@@ -38,37 +38,35 @@ private object LoginRateLimiter {
     private const val FREE_ATTEMPTS = 5
     private const val MINUTE_MS = 60_000L
 
-    // Precomputed Fibonacci sequence: FIB[n] = fib(n), 1-indexed (FIB[0] unused).
-    // Applied in minutes after FREE_ATTEMPTS consecutive failures.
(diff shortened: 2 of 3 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

PR #506, opened 2026-04-20
feat(install): add EXPOSE_LAN option and initial config writing

JordyPirata:
## Summary

- Adds `--expose-lan` CLI flag to `scripts/install.sh`
- Introduces `ambrosia_write_initial_config()` function that writes an initial `ambrosia.conf` on first install
- When `--expose-lan` is passed, sets `http-bind-ip=0.0.0.0`; otherwise defaults to `127.0.0.1` (localhost only)
- Skips config write if `~/.Ambrosia-POS/ambrosia.conf` already exists, preserving existing user configuration

## Motivation

By default, Ambrosia binds only to localhost. Users running the server on a dedicated machine (e.g., a Raspberry Pi or home server) need to accept connections from other devices on the local network without manually editing the config file after installation.

## Changes

| File | Change |
|------|--------|
| `scripts/install.sh` | Added `EXPOSE_LAN` flag, `ambrosia_write_initial_config()` function, and call within `ambrosia_install()` |

## Usage

```bash
# Default install (localhost only)
./install.sh

# Install with LAN exposure
./install.sh --expose-lan
```

Generated config (`~/.Ambrosia-POS/ambrosia.conf`):
```
http-bind-ip=0.0.0.0   # with --expose-lan
http-bind-ip=127.0.0.1 # default
http-bind-port=9154
```

## Test plan

- [ ] Run `./install.sh` on a clean machine — verify `ambrosia.conf` is created with `http-bind-ip=127.0.0.1`
- [ ] Run `./install.sh --expose-lan` — verify `ambrosia.conf` is created with `http-bind-ip=0.0.0.0`
- [ ] Run installer a second time on an existing install — verify `ambrosia.conf` is NOT overwritten
- [ ] Verify existing `--yes` and `--no-systemd` flags still work correctly

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -11,15 +11,20 @@ IFS=$'\n\t'
 # --- Argument validation ---
 AUTO_YES=false
 INSTALL_SYSTEMD=true
+EXPOSE_LAN=false
 
 for arg in "$@"; do
   case $arg in
-    --yes|-y) 
+    --yes|-y)
       AUTO_YES=true
       shift
       ;;
     --no-service)
       INSTALL_SYSTEMD=false
+      shift
+      ;;
+    --expose-lan)
+      EXPOSE_LAN=true
       shift
       ;;
     *)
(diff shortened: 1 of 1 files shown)


chrisguida:
Concept ACK, not tested yet

Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!

PR #504, opened 2026-04-16
Feat(WS): SSE payment proxy

Vidarte-Alberto:
This pull request refactors how WebSocket URLs are constructed and used in the client, simplifies the related configuration, and introduces a new API route for streaming WebSocket events via server-sent events (SSE). It also adds the `ws` package as a dependency and cleans up some platform-specific fields in `package-lock.json`.

**WebSocket and API configuration refactor:**

* Refactored `client/src/config/api.js` to remove the `getWsUrl` logic and related environment variable handling, simplifying the configuration to only export `API_URL`. The logic for constructing WebSocket URLs based on environment variables and browser location has been removed.
* Updated the test file `client/__tests__/api.config.test.js` to match the new configuration logic, now only testing `API_URL` and removing tests related to WebSocket URL construction.

**New API route for streaming payments events:**

* Added `client/src/app/api/ws-payments/route.js`, which creates a serverless API route that connects to the backend WebSocket using the `ws` package and streams messages to the client as server-sent events (SSE). This route handles connection, message forwarding, and cleanup on client disconnect.

**Dependency updates:**

* Added the `ws` package (`^8.18.0`) to `dependencies` in `client/package.json` and `client/package-lock.json` to enable server-side WebSocket connections. [[1]](diffhunk://#diff-1846122c2c83a486a3693f7966aa522c34cf489f674185c4da0d9221683fd81fL41-R42) [[2]](diffhunk://#diff-925cb809ce77c7f7f7f8bbd7594593438fc006dc56bf432a624c47722a784d3dL25-R26) [[3]](diffhunk://#diff-925cb809ce77c7f7f7f8bbd7594593438fc006dc56bf432a624c47722a784d3dL17100)

**Package lock cleanup:**

[...]

PR #507, opened 2026-04-21
fix(client): work behind reverse proxies with self-signed TLS

chrisguida:
## Summary

The Next.js middleware in `client/src/proxy.js` built its SSR `fetch()` URL as `new URL('/api/initial-setup', request.url)`. When the browser loads `https://host/`, `request.url` is the public HTTPS URL, so the middleware's SSR fetch tried to reach the app **via the reverse proxy** — and Node's undici fetch failed on the proxy's self-signed cert. In my setup (Caddy with `tls internal`) the error was:

```
TypeError: fetch failed
  [cause]: Error: error:0A0000C6:SSL routines:tls_get_more_records:packet length too long
  code: 'ERR_SSL_PACKET_LENGTH_TOO_LONG'
```

The `try/catch` swallowed the error, `initialized` stayed `null`, and the middleware fell through to the "no refresh token → /auth" branch — so new installs behind a reverse proxy **never reached the onboarding wizard**, even though the backend correctly reported `{initialized: false}`.

## Fix

Point SSR fetches at the **local Next.js process** directly via a new `INTERNAL_ORIGIN` constant:

```js
const INTERNAL_ORIGIN =
  process.env.INTERNAL_ORIGIN || `http://127.0.0.1:${process.env.PORT || 3000}`;
```

- **Docker**: Next.js binds to `0.0.0.0:3000` inside the container; `127.0.0.1:3000` from inside the same container still reaches it. No regression.
- **Electron**: `electron/utils/portAllocator.js` picks a free port in 3000–3100 and sets `PORT` in the spawned Next.js process's env (see `NextJsService.js:74`). `process.env.PORT || 3000` picks up the allocated port automatically.
- **Native** (`scripts/install.sh`): systemd unit doesn't override `PORT`, defaults to 3000.
- **`npm run dev`**: dev server on 3000. Works.
- **`INTERNAL_ORIGIN` env var** is an escape hatch for anyone binding Next.js to a non-loopback interface.

## Related — WebSockets behind a reverse proxy

I originally had a second patch in this PR for `getWsUrl()` in `config/api.js` — the WebSocket URL was hardcoded to `wss://host:9154` which didn't work behind a proxy (port 9154 isn't exposed, isn't TLS). Good news: I see #504 already refactored WebSocket handling to use a same-origin SSE route at `/api/ws-payments`, which works perfectly behind a proxy with no client-side changes. So that piece is already solved on `development`; this PR is now scoped to just the middleware fix.

## Reproducing / reviewer setup

If you want to reproduce the bug (and verify this PR fixes it), here's the exact Caddy config I used on Debian 12 (bookworm) in front of a native install.

**Install Caddy:**

```bash
sudo apt install -y debian-keyring debian-archive-keyring apt-transport-https curl
curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/gpg.key' | sudo gpg --dearmor -o /usr/share/keyrings/caddy-stable-archive-keyring.gpg
curl -1sLf 'https://dl.cloudsmith.io/public/caddy/stable/debian.deb.txt' | sudo tee /etc/apt/sources.list.d/caddy-stable.list
sudo apt update
sudo apt install -y caddy
```

**`/etc/caddy/Caddyfile`** (substitute your own hostname):

```
{
    local_certs
}

ambrosia.local {
    tls internal
    reverse_proxy localhost:3000
}
```

Then `sudo systemctl restart caddy`.

With `tls internal`, Caddy generates a local CA + leaf cert for the hostname, listens on 443 with a valid-but-untrusted TLS chain, and auto-redirects plain HTTP on port 80 to HTTPS. Everything goes through Next.js on port 3000, including the new `/api/ws-payments` SSE route — no separate `/ws/*` reverse_proxy needed.

**Also useful**, add a loopback entry so the machine can resolve its own `.local` name (Avahi doesn't answer queries for the machine's own hostname):

```
127.0.1.1 ambrosia.local
```

The middleware fix in this PR means Next.js no longer needs this to reach itself, but other internal tools (`curl`, health checks, etc.) still benefit.

## Test plan

- [x] Reproduced on a fresh install behind Caddy (`tls internal`) on `v0.6.0-beta` — wizard skipped
- [x] After patch applied to a rebuild of the client: wizard fires on first visit to `https://host/`
- [x] Plain-HTTP direct-to-Next.js install still works (no regression)
- [ ] Docker install (unchanged path — `fetch('http://127.0.0.1:3000/...')` from inside the `ambrosia-client` container is fine)
- [ ] Electron install (uses allocated `PORT` env — `INTERNAL_ORIGIN` picks it up automatically)

Context: hit this while building a fleet of OrangePi POS units behind per-unit Caddy for LAN-level HTTPS.

  Update (June 16, 2026)
  This PR has been rebased and expanded to include the following:
   - Security Fixes: Resolved high-severity vulnerabilities in Electron dependencies by implementing npm overrides in
     electron/package.json.
   - Code Cleanup: Removed explanatory comments in client/src/proxy.js as requested during review to maintain a cleaner
     codebase.
   - Rebase: Synchronized with the latest development branch, preserving the new CSP and nonce-based security logic.

### DIFF
--- a/client/src/proxy.js
+++ b/client/src/proxy.js
@@ -1,5 +1,8 @@
 import { NextResponse } from "next/server";
 
+const INTERNAL_ORIGIN =
+  process.env.INTERNAL_ORIGIN || `http://127.0.0.1:${process.env.PORT || 3000}`;
+
 export default async function proxy(request) {
   const { pathname } = new URL(request.url);
   const refreshToken = request.cookies.get("refreshToken");
@@ -20,7 +23,7 @@ export default async function proxy(request) {
   let needsBusinessType = false;
   try {
     const headers = { cookie: request.headers.get("cookie") || "" };
-    const initialSetupUrl = new URL("/api/initial-setup", request.url);
+    const initialSetupUrl = new URL("/api/initial-setup", INTERNAL_ORIGIN);
 
     const setupResponse = await fetch(initialSetupUrl, { headers });
     if (setupResponse.ok) {
@@ -56,7 +59,7 @@ export default async function proxy(request) {
   let shouldClearBusinessTypeCookie = false;
   try {
     const headers = { cookie: request.headers.get("cookie") || "" };
-    const configurationUrl = new URL("/api/config", request.url);
+    const configurationUrl = new URL("/api/config", INTERNAL_ORIGIN);
     const configResponse = await fetch(configurationUrl, { headers });
     if (configResponse.ok) {
       const configData = await configResponse.json();
@@ -95,6 +98,7 @@ export default async function proxy(request) {
     base-uri 'self';
     form-action 'self';
     frame-ancestors 'none';
+    upgrade-insecure-requests;
   `.replace(/\s{2,}/g, " ").trim();
 
   const requestHeaders = new Headers(request.headers);
--- a/electron/package-lock.json
+++ b/electron/package-lock.json
@@ -411,6 +411,7 @@
       "dev": true,
       "license": "BSD-2-Clause",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "cross-dirname": "^0.1.0",
         "debug": "^4.3.4",
@@ -432,6 +433,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "graceful-fs": "^4.2.0",
         "jsonfile": "^6.0.1",
@@ -448,6 +450,7 @@
       "dev": true,
       "license": "MIT",
       "optional": true,
+      "peer": true,
       "dependencies": {
         "universalify": "^2.0.0"
       },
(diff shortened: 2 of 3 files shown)


client/src/proxy.js
```diff
@@ -1,5 +1,14 @@
 import { NextResponse } from "next/server";
 
+// Middleware SSR fetches must target the local Next.js process directly,
```
Vidarte-Alberto: Please remove the comments to keep the code clean

Vidarte-Alberto (changes requested):

chrisguida:
Gonna rebase this and test it for real. Drafting for now

client/src/proxy.js
ca-ruz: Fixed

ca-ruz:
Built and tested locally. 

Screenshot From 2026-06-16 17-48-27

Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

chrisguida:
🚀

PR #532, opened 2026-04-30
Fix: Improve login error message when user role is deleted

### LINKED ISSUE #483: Incorrect error message when trying to log in with a user that has no role because it was deleted
I had a Cashier user with Cashier role and permissions, I deleted the Cashier role using the Admin user, then when I tried to log in as the Cashier user again I get this error message: 

Image

but it should say something like: "No assigned role for this user, contact Admin"

ca-ruz:
This PR addresses issue #483 by providing a more helpful error message when a user attempts to log in but their assigned role has been deleted. Previously, the system would return a generic "Invalid credentials" error, which misled users into thinking they had entered an incorrect PIN.

  Backend:
   - Exceptions: Added MissingRoleException to handle cases where a valid user lacks an active role.
   - AuthService: Updated the authentication query to use a LEFT JOIN. This allows the system to find the user even if their role_id is null or points to a deleted role.
   - Logic: Implemented a check that throws MissingRoleException only after successful PIN verification, ensuring security is maintained.
   - Error Handling: Updated Handler.kt to map this exception to an HTTP 401 response with the message: "No assigned role for this user, contact Admin".
   - API Cleanup: Refactored the /login route in Authorize.kt to correctly propagate exceptions to the global error handler.

  Frontend:
   - UI Logic: Modified PinLogin.jsx to detect the specific error message from the backend and display it directly to the user.
   - Fallback: Maintained the generic "Incorrect PIN" message for all other authentication failures to prevent username enumeration.

  Testing:
   - E2E: Added a regression test test_login_fails_when_role_deleted in test_auth_e2e.py to verify the full API flow.
   - Unit: Added a test case in PinLogin.test.jsx to ensure the React component correctly renders the specific error message.

  Fixes #483 
  
Screenshot From 2026-04-30 14-51-49


client/src/components/pages/Auth/PinLogin/__tests__/PinLogin.test.jsx
```diff
@@ -120,4 +120,24 @@ describe("PinLogin", () => {
 
     expect(screen.queryByText(/lockout\.message/)).not.toBeInTheDocument();
   });
+
+  it("shows the specific error message when the user's role is deleted", async () => {
+    const specificMessage = "No assigned role for this user, contact Admin";
+    const err = new Error(specificMessage);
```
Vidarte-Alberto: Please use "error" instead of the abbreviation. We should use more descriptive variable names here.

client/src/components/pages/Auth/PinLogin/PinLogin.jsx
```diff
@@ -108,6 +108,8 @@ export default function PinLogin() {
         setLockedUntil(ts);
         localStorage.setItem("pinLockoutUntil", ts.toString());
         setError("");
+      } else if (err?.message === "No assigned role for this user, contact Admin") {
```
Vidarte-Alberto: Please use "error" instead of the abbreviation. We should use more descriptive variable names here.

Vidarte-Alberto (changes requested):
Do fixes

client/src/components/pages/Auth/PinLogin/__tests__/PinLogin.test.jsx
ca-ruz: Fixed!

client/src/components/pages/Auth/PinLogin/PinLogin.jsx
ca-ruz: Fixed!

ca-ruz:
Fixed variable name, built and tested, still working:

Screenshot From 2026-04-30 14-51-49

Vidarte-Alberto (approved):
LGTM!!!

Sharmaz (approved):
LGTM!!

PR #559, opened 2026-05-18
fix silent refresh error and /auth/refresh race condition. Remove isL…

### LINKED ISSUE #523: Nav items intermittently disappear from the sidebar/navbar
## Summary

Navigation items (pages) occasionally vanish from the sidebar and/or bottom nav bar during normal use. The nav renders momentarily empty and then either repopulates or stays blank, forcing the user to reload.

## Affected components

| File | Role |
|---|---|
| `client/src/components/ModuleNavigation.js` | Restaurant sidebar |
| `client/src/components/pages/Store/StoreLayout/Sidebar.jsx` | Store sidebar |
| `client/src/components/pages/Store/StoreLayout/BottomNav.jsx` | Mobile bottom nav |
| `client/src/components/navigation/DynamicNavigation.jsx` | Generic nav |
| `client/src/hooks/useModules.jsx` | Navigation data source |
| `client/src/lib/modules.js` | `getAvailableNavigation` / `getAvailableModules` |

## Steps to reproduce

1. Log in with any role.
2. Let the session reach the silent-refresh window (access token TTL ~15 min by default) **or** open the app on a slow network.
3. Observe the sidebar blank.

Sharmaz:
## Changes:
- Fix nav items disappearing during use: silent refresh errors (`focus`/`visibilitychange`) no longer reset auth state — a new `SILENT_REFRESH_ERROR` action preserves `isAuth` and `permissions` on transient network failures
- Fix race condition on token expiry: concurrent requests sharing an expired token now share a single `/auth/refresh` call via a module-level promise mutex, preventing a second refresh attempt from dispatching a false `auth:expired` event
- Remove redundant `isLoading` guard from `useNavigation` that was blanking the nav unnecessarily when `isLoading` and `isAuth` were both true simultaneously

**Related Issues**:
https://github.com/olympus-btc/ambrosia/issues/523

### DIFF
--- a/client/src/hooks/useNavigation.jsx
+++ b/client/src/hooks/useNavigation.jsx
@@ -12,15 +12,9 @@ export function useNavigation() {
   const isAdmin = user?.isAdmin || false;
   const { businessType } = useConfigurations();
 
-  const availableFeatures = useMemo(() => {
-    if (isLoading) return {};
-    return getAvailableFeatures(isAuth, isAdmin, permissions, businessType);
-  }, [isAuth, isAdmin, isLoading, permissions, businessType]);
+  const availableFeatures = useMemo(() => getAvailableFeatures(isAuth, isAdmin, permissions, businessType), [isAuth, isAdmin, permissions, businessType]);
 
-  const availableNavigation = useMemo(() => {
-    if (isLoading) return [];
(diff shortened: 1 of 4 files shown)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

Built and tested locally, everything works as expected.

PR #563, opened 2026-05-20
Fix/shift amount

### LINKED ISSUE #536: Bug: Active Shift Widget does not sync with new sales
## Summary

  When a sale is completed while a shift is open, the active shift widget continues to show Total sales = $0.00
  and Total tickets = 0 instead of reflecting the new transaction. This makes the shift summary unreliable for
  cash control and end-of-shift reconciliation.

  The current client implementation derives shift totals from /tickets, but the filtering logic and refresh
  behavior are not aligned with how ticket_date is actually stored and returned by the backend. As a result, an
  open shift can remain unchanged even after new tickets are created.

  ## Current behavior

  - The active shift widget renders totalSales and totalTickets through client/src/hooks/turn/useShiftTickets.js,
    not from persisted backend shift totals.
  - useShiftTickets filters tickets with Number(t.ticket_date) >= shiftStartMs, which assumes ticket_date is a
    numeric timestamp.
  - The backend defines ticket_date as a String in server/app/src/main/kotlin/pos/ambrosia/models/
    AppModels.kt:219, and the database stores it as TEXT with datetime('now') in server/app/src/main/resources/
    db/migration/V1__db-schema.sql:145.
  - Tickets returned by server/app/src/main/kotlin/pos/ambrosia/server/app/src/main/kotlin/pos/ambrosia/api/
    Tickets.kt come from server/app/src/main/kotlin/pos/ambrosia/services/TicketService.kt, which exposes
    ticket_date as stored rather than converting it to epoch time.
  - The hook only refetches when shift_date or start_time changes, so creating a new sale does not automatically
    refresh the widget.
  - The current test coverage in client/src/hooks/turn/__tests__/useShiftTickets.test.js uses mocked ticket_date
    values as numeric timestamp strings, which does not match the actual backend format.

  ## Expected behavior

  - When a sale is completed during an open shift, the active shift widget should immediately reflect the updated
    Total sales and Total tickets.
  - Shift ticket filtering should work with the real ticket_date format returned by the backend.
  - The active shift metrics should refresh after a sale or ticket is created, without requiring the user to
    manually collapse or reopen the widget.
  - Tests should cover the real ticket_date format and verify shift updates after new sales are recorded.

  ## Files to investigate

  | Area | Why it matters |
  | --- | --- |
  | client/src/hooks/turn/useShiftTickets.js | This is where totalBalance and totalTickets are computed; the
  current filter depends on Number(t.ticket_date). |
  | client/src/components/turn/ShiftWidget.jsx | The widget consumes the hook and currently only loads ticket
  metrics when expanded. |
  | client/src/modules/orders/ordersService.js | Exposes getTickets() and the sale/ticket creation flow that
  should trigger a shift refresh. |
  | server/app/src/main/kotlin/pos/ambrosia/services/TicketService.kt | Confirms the actual ticket_date format
  sent to the client. |
  | server/app/src/main/resources/db/migration/V1__db-schema.sql | Defines ticket_date as TEXT with
  datetime('now'), which is the root of the current mismatch. |
  | client/src/hooks/turn/__tests__/useShiftTickets.test.js | Current tests validate a date format that does not
  match backend behavior. |

  ## Implementation notes

  - The most likely root cause is a format mismatch between what the client expects and how ticket_date is
    actually stored and returned by the backend.
  - The date comparison should be normalized to one consistent format, either by parsing the real backend date
    string correctly in the client or by exposing a stable explicit timestamp format from the backend.
  - In addition to the date parsing issue, the widget needs a refresh strategy after new sales are created: data
    invalidation, explicit refetch, shared eventing, or centralized state.
  - If shift totals are expected to be reliable for opening and closing cash control, it may be worth evaluating
    whether these values should come from a dedicated shifts endpoint instead of being recomputed only in the UI.
  - Tests should stop assuming numeric timestamp strings and should cover values such as YYYY-MM-DD HH:mm:ss, or
    whatever exact format the API returns.

  ## Acceptance criteria

  - [ ] With an open shift, creating a new sale updates the active shift widget with the correct Total sales and
    Total tickets.
  - [ ] Shift sales/ticket calculations work with the real ticket_date format returned by the backend.
  - [ ] The widget does not remain stale after the initial load and reflects new sales without manually reopening
    the shift view.
  - [ ] At least one test covers backend-realistic ticket date values and verifies they are counted within the
    active shift.
  - [ ] There is coverage for refreshing active shift metrics after a new sale or ticket is created.

## Screenshots

Image

Sharmaz:
## Changes:
- Fix shift widget showing $0.00 and 0 tickets: `ticket_date` from SQLite `datetime('now')` is UTC with space separator — parse it as UTC with `.replace(" ", "T") + "Z"` instead of comparing as a number (`NaN`) or as local time (wrong epoch)
- Fix shift totals including tickets from previous shifts: the UTC/local timezone mismatch caused old tickets to appear in the future relative to the new shift start
- Replace global `ticket:created` window event with direct `refreshShiftTickets()` call via `useTurn()` — explicit React data flow with no implicit coupling
- Extract `useShiftTicketMetrics` hook from `TurnProvider`, separating ticket metrics concern from shift lifecycle; delete dead `useShiftTickets` hook
- Move `TurnProvider` below `I18nProvider` in the provider hierarchy to allow `useTranslations` inside the provider
- Fix abbreviations across the turn module: `t` → `shiftTranslations`, `err` → `caughtError`, `p`/`m` → `payment`/`method`
- Add `useShiftTicketMetrics` tests covering UTC date filtering, previous-shift exclusion, breakdown, and `reset()`

**Related Issues**:
https://github.com/olympus-btc/ambrosia/issues/536


Vidarte-Alberto (approved):
LGTM!!

JordyPirata (approved):
LGTM!!

Issue #564, opened 2026-05-20
Blank product SKU is stored as an empty string and triggers duplicate SKU errors


ca-ruz:
## Summary
  When creating or updating a product without a SKU, the app should treat the field as empty and allow multiple products to be saved without a SKU.

  ## Actual behavior
  If I create one product with the SKU field left blank, it saves successfully. If I then create a second product and also leave SKU blank, the request fails with a `SKU already exists` error.

  ## Expected behavior
  Blank or whitespace-only SKU values should be treated as `null` and should bypass the SKU uniqueness check. Products without SKUs should be able to coexist.

  ## Notes
  - This appears to happen because blank SKU values are still being sent/stored as `""` instead of `null`.
  - The duplicate SKU check should only run when the SKU is a real non-empty value.
  - This affects the product create/update flow in the Store UI.

<img width="1026" height="606" alt="Image" src="https://github.com/user-attachments/assets/867171ad-c4bf-4d82-b1a6-e2fee93862de" />

ca-ruz:
Addressed in PR: #565

PR #565, opened 2026-05-21
Fix: Treat blank product SKU as null

ca-ruz:
- Normalize blank or whitespace-only product SKUs to `null` on the backend before validation and persistence
  - Skip the duplicate-SKU lookup when SKU is blank, so products without SKUs can be created repeatedly
  - Update the client product form payload to send `SKU: null` instead of an empty string when the field is blank
  - Add regression coverage for both the backend service and the client request payload

Closes the blank-SKU duplicate error bug: #564 

Screenshot From 2026-05-19 12-10-44


client/src/components/pages/Store/hooks/useProducts.jsx
```diff
@@ -17,6 +17,11 @@ export function useProducts() {
   const [error, setError] = useState(null);
   const { upload, isUploading } = useUpload();
 
+  const normalizeSku = (sku) => {
+    const trimmed = typeof sku === "string" ? sku.trim() : sku;
+    return trimmed ? trimmed : null;
+  };
+
```
Sharmaz: You can use this instead:
`const normalizeSku = (sku) => sku?.trim() || null;`

ca-ruz: great! I didn't know that one liner did the same, thanks.

JordyPirata (approved):
LGTM!!

Sharmaz (approved):
LGTM!!

PR #568, opened 2026-05-21
fix(client): solve step, and maxValue limits

### LINKED ISSUE #552: Bug: Decimal Rounding in Cash Payment Modal Input
In the cash payment modal (`CashPaymentModal`), the received amount field rounds values to 1 decimal place when displaying them. For example, typing `0.69` shows `0.7` in the input.

The **value sent to the server is correct** (0.69), so the change calculation is visually wrong but does not affect the final transaction. However, it confuses the cashier and can undermine trust in the system.

## Steps to Reproduce

1. Go to the cart with at least one product
2. Select "Cash" as the payment method
3. In the modal, manually type a value with 2 decimal places such as `0.69`
4. Observe that the field shows `0.7` instead of `0.69`

## Expected Behavior

The input should display the value with 2 decimal places exactly as the user types it: `0.69`.

## Actual Behavior

The input rounds to 1 decimal place: `0.7`.

## Affected File

`client/src/components/pages/Store/Cart/CashPaymentModal.jsx` (lines 84–104)

## Root Cause

HeroUI's `NumberInput` has `step={0.10}`, which causes the component to restrict display precision to the number of decimals in the step value (1 decimal). The internal value is correct but the display is truncated.

```jsx
<NumberInput
  value={cashReceived}
  step={0.10}    // <-- causes visual rounding to 1 decimal
  minValue={0}
  ...
/>
```

## Suggested Fix

Change `step` from `0.10` to `0.01` so the input allows and displays 2 decimal places:

```jsx
<NumberInput
  value={cashReceived}
  step={0.01}    // allows 2-decimal display
  minValue={0}
  ...
/>
```

If the increment/decrement buttons (+/-) should still move in 10-cent steps, `step={0.01}` still works fine since users type values freely.

### LINKED ISSUE #553: Bug: Cash Modal Does Not Allow Amounts Greater Than 4 Digits
## Description

In the cash payment modal (`CashPaymentModal`), it is not possible to enter an amount greater than 4 digits (i.e., greater than 9999). The input stops accepting characters or truncates the value when that limit is reached.

This prevents recording high-value cash payments (e.g., $10,000 or more), which is a real scenario for large sales or regions with inflation where bills have high denominations.

## Steps to Reproduce

1. Have a high-value product in the cart (or open the cash modal directly)
2. Try to type an amount greater than 4 digits, e.g. `10000`
3. Observe that the input does not accept the 5th digit or the value gets truncated

## Expected Behavior

The field should accept any reasonable amount (e.g., up to 6–7 digits: $999,999).

## Actual Behavior

The input limits entry to 4 digits (maximum ~9999).

## Affected File

`client/src/components/pages/Store/Cart/CashPaymentModal.jsx` (lines 84–104)

## Probable Root Cause

HeroUI's `NumberInput` has an internal digit limit (possibly inherited from `react-aria`), or the `onChange` handler using `parseFloat` loses precision for large values. There is no explicit `maxValue` in the code, which suggests an implicit ceiling from the component itself.

```jsx
<NumberInput
  value={cashReceived}
  minValue={0}
  step={0.10}
  // No maxValue → implicit limit from the component
  onChange={(e) => {
    setCashReceived(parseFloat(e.target.value) || 0);
  }}
/>
```

## Suggested Fix

**Option 1:** Add an explicit `maxValue` large enough:

```jsx
<NumberInput
  value={cashReceived}
  minValue={0}
  maxValue={9999999}
  step={0.01}
  ...
/>
```

**Option 2:** Replace `NumberInput` with a controlled `<input type="number">` that has no internal component restrictions:

```jsx
<input
  type="number"
  min={0}
  step={0.01}
  value={cashReceived}
  onChange={(e) => setCashReceived(parseFloat(e.target.value) || 0)}
/>
```

This option gives full control but loses HeroUI styling; CSS classes would need to be applied manually.

## Notes

Check HeroUI's `NumberInput` for a `maxLength` prop or similar that may be applied by default. Also verify whether this issue is related to the decimal rounding bug (Issue 2), since both originate in the same component and the `step` value may affect internal validation.

Vidarte-Alberto:
This pull request updates the cash input handling in the `CashPaymentModal` to improve input validation, enforce a maximum allowed value, and enhance formatting for user experience. The changes ensure that users cannot enter values above 9,999,999, restrict input to valid numbers, and format the input with grouping and two decimal places.

**Input validation and value enforcement:**
- Updated the `onChange` handler to sanitize input by removing non-numeric characters, parse the value, and cap it at 9,999,999.
- Added an `onKeyDown` handler to prevent entering digits that would cause the value to exceed 9,999,999.
- Set the `maxValue` prop to 9,999,999 to enforce the upper limit at the component level.

**Formatting improvements:**
- Added `formatOptions` to display the input with digit grouping and a maximum of two decimal places.
- Changed the input step value from 0.10 to 0.01 for finer increments.

**Code clarity:**
- Renamed the `onValueChange` parameter from `value` to `receivedAmount` for clarity.

**Screenshots**
image
image

**Related Issues**
Resolves #553
Resolves #552

### DIFF
--- a/client/src/components/pages/Store/Cart/CashPaymentModal.jsx
+++ b/client/src/components/pages/Store/Cart/CashPaymentModal.jsx
@@ -84,18 +84,32 @@ export function CashPaymentModal({
           <NumberInput
             label={t("receivedLabel")}
             value={cashReceived}
-            onValueChange={(value) => {
-              setCashReceived(value ?? 0);
+            onValueChange={(receivedAmount) => {
+              setCashReceived(receivedAmount ?? 0);
               setError("");
             }}
             onChange={(e) => {
               if (e?.target) {
-                setCashReceived(parseFloat(e.target.value) || 0);
+                const parsedAmount = parseFloat(e.target.value.replace(/[^0-9.]/g, "")) || 0;
+                setCashReceived(Math.min(parsedAmount, 9999999));
                 setError("");
               }
             }}
+            onKeyDown={(e) => {
+              if (!/^[0-9]$/.test(e.key)) return;
+              const currentInputText = (e.target.value || "").replace(/[^0-9.]/g, "");
+              if (parseFloat(currentInputText + e.key) > 9999999) {
+                e.preventDefault();
+              }
+            }}
             minValue={0}
-            step={0.10}
+            maxValue={9999999}
+            formatOptions={{
+              useGrouping: true,
+              maximumFractionDigits: 2,
+            }}
+
+            step={0.01}
             size="lg"
             classNames={{ inputWrapper: "shadow-none" }}
             startContent={


Sharmaz (approved):
LGTM!!

JordyPirata (approved):
LGTM!!

PR #570, opened 2026-05-22
Feat(hardware): implement rebuildable image for Raspberry Pi Zero 2W and Orange Pi Zero 2W

Vidarte-Alberto:
This PR introduces a fully automated, reproducible SD card image build system for Ambrosia POS on ARM hardware, building on the manual device setup work pioneered by @chrisguida in #544.

  Background

  #544 established the foundational hardware deployment path: a manual process for provisioning OrangePi Zero 2W devices with the full Ambrosia stack — server, client, Phoenixd, Caddy, and the Wi-Fi setup portal. That work defined the
   runtime architecture, service layout, and first-boot flow that this PR now automates end-to-end.

  Reproducible image build system

  - New hardware/image/ tree with a board-agnostic assembler (assemble-image.sh) that builds flashable SD card images from a clean base OS, installs all runtime dependencies inside a chroot, and emits a compressed image + SHA-256 +
  build manifest
  - Full stack embedded in the image: Ambrosia server (Kotlin/Ktor), Next.js client, Phoenixd, Caddy, NetworkManager, Avahi, hostapd, dnsmasq, openssh
  - build-artifacts.sh separates the Gradle/npm compile step from the image assembly so artifacts can be cached and reused across builds (--skip-artifacts-build)
  - Build produces ambrosia-<board-id>-<version>.img.gz + .sha256 + .manifest.json with full build metadata (versions, commit SHA, base image provenance)

  Multi-board support

  - Board-agnostic assembler: all board-specific config lives in hardware/image/boards/<board-id>/board.env
  - OrangePi Zero 2W (opi-zero-2w) — Debian Bookworm official OPi image
  - Raspberry Pi Zero 2W (rpi-zero-2w) — Raspberry Pi OS Lite Bookworm 64-bit
  - Adding a new board requires only a board.env, packages.txt, and README.md

  First boot & preseed

  - ambrosia-firstboot service runs once on first boot: regenerates SSH host keys and machine ID, applies a preseed file from the boot partition (ambrosia-device.env), sets hostname, locale, Wi-Fi country, and admin password
  - Hostname is auto-generated from board short name + machine ID prefix (ambrosia-rpi-XXXXXX, ambrosia-opi-XXXXXX)
  - Boot path detection handles both RPi OS (/boot/firmware/) and OPi (/boot/) automatically

  Wi-Fi setup portal

  - On first boot without a known Wi-Fi network, the device brings up a captive-portal hotspot (<hostname>-setup) where the operator enters SSID + password
  - Portal forces wlan0 managed and radio on at startup — works unattended on RPi OS Bookworm without manual intervention
  - SSID scan runs before AP mode so the portal shows nearby networks
  - Background monitor exits the portal cleanly when WiFi is connected externally, restoring Caddy via systemd ExecStopPost

  RPi-specific fixes

  - Suppresses the RPi OS Bookworm interactive "rename user" first-boot wizard by writing userconf.txt to the boot partition during image assembly
  - Fallback partition detection when lsblk/blkid fail to report FSTYPE on loop devices (assumes last partition = root, first = boot)


JordyPirata:
**No update path for the Ambrosia artifacts.**

This build bakes `ambrosia.jar` and the Next.js client into `/opt/ambrosia/` at image-assembly time (`install_ambrosia_artifacts()` in `hardware/image/build/assemble-image.sh`), but there is currently no mechanism to update them on a running device. Once a unit is flashed and deployed, the only ways to get a new version onto it are:

- re-flashing the SD card, which wipes device state (database, Phoenix seed, config), or
- manually replacing the files over SSH / writing them directly to the filesystem, with no tooling, no version tracking, and no rollback.

Not a blocker for this PR, but worth capturing as a follow-up: an upgrade script (or `ambrosia-update` service) that swaps the artifacts in `/opt/ambrosia/`, restarts the services, and preserves `/home/ambrosia/.Ambrosia-POS` and `/home/ambrosia/.phoenix`. The manifest already records version + SHA-256 per build, so the verification half of that mechanism exists.

hardware/image/boards/opi-zero-2w/board.conf
```diff
@@ -0,0 +1,16 @@
+BOARD_ID="opi-zero-2w"
```
Sharmaz, line 1: Why you are filtering a .env file?

hardware/image/boards/rpi-zero-2w/board.conf
```diff
@@ -0,0 +1,16 @@
+BOARD_ID="rpi-zero-2w"
```
Sharmaz, line 1: Why you are filtering a .env file?

Sharmaz (changes requested):
Open to discussion

Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #572, opened 2026-05-25
fix(store):  Cart persistent state

JordyPirata:
## Summary

- Exposes a `hydrated` flag from `usePersistentCart` to signal when the cart has finished rehydrating from `localStorage`
- Disables the Pay button until `hydrated === true` to prevent payments before the cart state is ready
- Adds a reconciliation effect in `Cart` that, once hydrated and products are loaded, updates stale prices from `localStorage` and removes products that no longer exist

## Test plan

- [x] Add a product to the cart, reload the page — cart should persist and Pay should remain disabled until hydration completes
- [x] Change the price of a product in the backend while a cart is saved in `localStorage`, reload — the cart should update to the new price automatically
- [x] Add a product, delete it from the backend, reload — the product should be removed from the cart on load
- [x] All 17 Cart test suites pass (`npx jest --testPathPatterns="Cart|usePersistentCart|SummaryContent"`)
- [x] Lint passes (`npm run lint`)

### DIFF
--- a/client/src/components/pages/Store/Cart/Cart.jsx
+++ b/client/src/components/pages/Store/Cart/Cart.jsx
@@ -22,10 +22,30 @@ export function Cart() {
     cart,
     setCart,
     discount,
+    hydrated,
     resetCartState,
   } = usePersistentCart();
   const { products, refetch: refetchProducts } = useProducts();
   const { categories } = useCategories();
+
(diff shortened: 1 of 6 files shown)


Sharmaz (approved):

Vidarte-Alberto (approved):
LGTM!!

PR #593, opened 2026-06-09
Chore/client security

Sharmaz:
## Changes:
  - Remove wildcard CORS headers from API proxy routes in `route.js` and `next.config.mjs`; Next.js API routes are same-origin by default and do not require CORS headers
  - Add non-CSP security headers to all routes: `X-Content-Type-Options`, `X-Frame-Options`, `Referrer-Policy`, `Permissions-Policy`
  - Add nonce-based Content Security Policy via `proxy.js` middleware; nonce is generated per-request and passed to Next.js for automatic application to hydration scripts
  - Add `https://api.coingecko.com` to `connect-src`; fix `style-src-elem` to use `'unsafe-inline'` in development only (restores Next.js DevTools widget without weakening production CSP)
  - Normalize product image URLs in `ImageUploader` with `storedAssetUrl()` to fix `img-src` CSP violation caused by absolute backend URLs

### DIFF
--- a/client/src/proxy.js
+++ b/client/src/proxy.js
@@ -86,11 +86,11 @@ export default async function proxy(request) {
   const csp = `
     default-src 'self';
     script-src 'self' 'nonce-${nonce}' 'strict-dynamic'${isDev ? " 'unsafe-eval'" : ""};
-    style-src-elem 'self' 'nonce-${nonce}';
+    style-src-elem 'self' ${isDev ? "'unsafe-inline'" : `'nonce-${nonce}'`};
     style-src-attr 'unsafe-inline';
     img-src 'self' blob: data:;
     font-src 'self';
-    connect-src 'self' ws: wss:;
+    connect-src 'self' ws: wss: https://api.coingecko.com;
     object-src 'none';
     base-uri 'self';
     form-action 'self';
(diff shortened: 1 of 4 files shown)


ca-ruz (approved):
LGTM! 

Built and tested locally, on both mobile and desktop, works as expected.

client/src/proxy.js
```diff
@@ -80,7 +80,31 @@ export default async function proxy(request) {
     }
   }
 
-  const next = NextResponse.next();
+  const nonce = Buffer.from(crypto.randomUUID()).toString("base64");
+  const isDev = process.env.NODE_ENV === "development";
+
+  const csp = `
```
Vidarte-Alberto: Please use more descriptive names

Vidarte-Alberto (changes requested):
Please do fixes

client/src/proxy.js
Sharmaz: I know that you dont have enought context on this let me enlighten you: 
- nonce: https://developer.mozilla.org/en-US/docs/Glossary/Nonce
- isDev: isDevelopment boolean variable https://dev.to/michi/tips-on-naming-boolean-variables-cleaner-code-35ig
- csp: ContentSecurityPolicy https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP

Vidarte-Alberto (approved):
LGTM!!

PR #599, opened 2026-06-17
Fix/wallet access

Sharmaz:
## Changes:
- Fix `loginWallet` to reject on non-2xx responses, closing a bypass where any password (even an incorrect one) granted access to the Wallet guard.
- Add real server-side revocation for `walletAccessToken`: persist it in a new `wallet_token` column on `users`, validate it against the database on every wallet-scoped request, and reduce its TTL from 8h to 5 minutes.
- Revoke the wallet token on `/wallet/logout` instead of only clearing the cookie, so navigating away from the Wallet/Seed screens actually invalidates server-side access.
- Add `TokenServiceTest` covering wallet token generation, validation, and revocation.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -95,13 +95,13 @@
       "license": "ISC"
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.29.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.0.tgz",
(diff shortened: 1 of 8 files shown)


ca-ruz (approved):
LGTM!

Built and tested locally, on both mobile and laptop. Works as expected.

Vidarte-Alberto (approved):
LGTM!

PR #603, opened 2026-06-18
fix(client): render payment modals once in Cart to fix duplicate modal on mobile

Vidarte-Alberto:
## Problem

  On mobile, after a Lightning payment is confirmed, the QR code reappears as if waiting for payment again.

  **Root cause:** `SummaryContent` rendered `BitcoinPaymentModal`, `CashPaymentModal`, and `CardPaymentModal` inline. Since `SummaryContent` is used in two
  places — the desktop sidebar (`Summary`) and the mobile modal (`SummaryModal`) — there were always **two simultaneous instances** of each payment modal
  mounted in the DOM, each generating its own independent Lightning invoice.

  HeroUI `Modal` renders via a portal at the document root, bypassing the `hidden md:block` CSS on the desktop sidebar. This means both modal instances were
  **visually active on mobile** at the same time.

  When the user paid the mobile instance's invoice:
  1. The mobile modal showed "confirmed" and closed (cart reset → `SummaryModal` closed → modal unmounted).
  2. The desktop instance's portal — which had its own unpaid invoice and `paymentReceived = false` — remained open and became visible, showing a new QR
  waiting for payment.

  ## Fix

  Moved the three payment modals out of `SummaryContent` and rendered them **once** at the `Cart` level, outside the desktop/mobile split.

### DIFF
--- a/client/src/components/pages/Store/Cart/Cart.jsx
+++ b/client/src/components/pages/Store/Cart/Cart.jsx
@@ -8,6 +8,9 @@ import { PageHeader } from "@/components/shared/PageHeader";
 import { useCategories } from "../hooks/useCategories";
 import { useProducts } from "../hooks/useProducts";
 
+import { BitcoinPaymentModal } from "./BitcoinPaymentModal";
+import { CardPaymentModal } from "./CardPaymentModal";
+import { CashPaymentModal } from "./CashPaymentModal";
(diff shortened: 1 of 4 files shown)


Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM! 

Built and tested locally on both mobile and laptop. Works as expected.

PR #607, opened 2026-06-19
remove deps overrides from package.json.

Sharmaz:
## Changes:
- Remove unnecessary npm `overrides` for `brace-expansion`, `form-data`, `joi`, `js-yaml`, `tar`, and `undici` in `electron/package.json` — each dependency's own declared version range already resolves to the same (audit-clean) version without forcing it, so the overrides were dead weight.

### DIFF
--- a/electron/package-lock.json
+++ b/electron/package-lock.json
@@ -350,6 +350,23 @@
         "node": ">=16.4"
       }
     },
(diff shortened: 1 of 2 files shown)


ca-ruz (approved):
LGTM!

Built and tested locally. Works as expected.

Vidarte-Alberto (approved):
LGTM!!

PR #609, opened 2026-06-21
fix(store): unblock re-adding items pending removal

JordyPirata:
## Summary

When an item is removed from the cart it enters a 5s "pending removal" window (hidden but still in `cart`) so it can be undone via a toast. During that window, tapping the same product in the catalog incremented the **hidden** item's quantity, so it never reappeared until the timeout deleted it — making it feel like you couldn't re-add the item for 5 seconds.

## Changes

- **`usePendingRemoval.js`** — the hook now owns the toast lifecycle. `startRemoval(id, onConfirm, toastKey)` stores the toast key, and `cancelRemoval`, the natural 5s timeout, and `clearPendingRemovals` all dismiss the toast via `closeToast`. Cancelling now also cancels the timeout and removes the toast.
- **`SummaryContent.jsx`** — creates the toast, gets its key, and passes it to `startRemoval`. The "undo" button just calls `cancelRemoval(id)` (toast dismissal is the hook's job). Removed the now-unneeded local `useRef`/`closeToast` handling.
- **`Cart.jsx`** — new `handleAddProduct`: if the product is pending removal, it cancels the pending removal (restoring the item and dismissing its toast) instead of staying blocked. Wired into `SearchProducts`.

## Behavior note

Re-tapping a product that is mid-removal **restores it as-is** (undoes the removal) rather than adding +1. For a single-quantity item this yields exactly 1, which matches the natural expectation.

## Tests

- `usePendingRemoval.test.js` — toast is closed on cancel, on confirm after 5s, and on clear.
- `SummaryContent.test.jsx` — verifies the toast key is passed to `startRemoval`.

All 179 Cart tests pass; lint clean on the touched files.

### DIFF
--- a/client/src/components/pages/Store/Cart/Cart.jsx
+++ b/client/src/components/pages/Store/Cart/Cart.jsx
@@ -1,5 +1,5 @@
 "use client";
-import { useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 
 import { useTranslations } from "next-intl";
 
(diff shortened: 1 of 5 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

PR #612, opened 2026-06-21
fix force dynamic rendering on root page avoiding scripts blocked by CSP.

Sharmaz:
Changes:
- Split `app/page.jsx` into a Server Component wrapper and a new `Home.jsx` Client Component, fixing a bug where the root route was statically prerendered at build time — causing the CSP nonce (regenerated per request by the middleware) to never match the build-time HTML, blocking every script and leaving the app stuck on the loading screen.

### DIFF
--- a/client/src/app/page.jsx
+++ b/client/src/app/page.jsx
@@ -1,45 +1,7 @@
-"use client";
-
-import { useEffect, useRef } from "react";
-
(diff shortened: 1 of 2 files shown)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #623, opened 2026-06-26
Fix: Allow currency search by currency name

ca-ruz:
- Allow currency autocomplete search to match currency names, not only currency codes.
  - Add a shared `CurrencyInput` component for the repeated currency picker logic.
  - Use the shared component in both Settings and Onboarding.
  - Add focused regression coverage for searching by currency name, for example `mex` matching `MXN - Mexican Peso`.

Screenshot From 2026-06-25 20-10-29

Screenshot From 2026-06-25 20-10-45

### DIFF
--- a/client/src/components/pages/Onboarding/AddBusinessData.jsx
+++ b/client/src/components/pages/Onboarding/AddBusinessData.jsx
@@ -2,9 +2,10 @@
 
 import { useState, useMemo } from "react";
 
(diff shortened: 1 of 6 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!!

PR #627, opened 2026-06-27
Feature/discounts

### LINKED ISSUE #515: Item discounts
There is currently no way to apply a discount to specific items in an order. During events, promotions, or special occasions, merchants need to offer percentage-based discounts on selected products (e.g., 20% off appetizers during happy hour, discounted drinks at a special event) without manually adjusting item prices each time.
                                                                                        
Allow merchants to apply a percentage discount to individual items in the Cart / Order Summary:
- Tap an item in the cart to apply a percentage discount (e.g., 10%, 20%, 50%).                                                                                                                                                     
- The discounted price should be shown clearly next to the original price on the item line.                                                                                                                                         
- The order total should update automatically to reflect the discount.                                                                                                                                                              
- Discounts should be per-item, not applied globally to the whole order.                                                                                                                                                            
                                                                                                                                                                                                                                      
Optionally, allow saving named discount presets (e.g., "Event Night -20%") so the same discount can be quickly reapplied across multiple items or future orders.
                                                                                                                                                                    
- Manually editing item prices before the event — error-prone and requires reverting prices afterward.
- Applying a global discount to the full order — too broad when only specific items are discounted.                                                                                                                                 

**Additional context**                                                                                                                                                                                                              
Merchants mentioned this need specifically in the context of events, where a subset of menu items is offered at a reduced price for a limited time. Percentage-based discounts are preferred over fixed amounts as they scale naturally across items with different prices.

Sharmaz:
## Changes:
- Add manual discount to store orders: cashiers can apply a fixed or percentage discount from the cart summary, with real-time total preview while typing
- Persist discount type (fixed/percentage) in localStorage alongside discount value
- Send discount amount to the server on checkout; persist it on the order record with a new DB migration
- Fix UI: remove shadow from NumberInput fields (Quantity and Discount) to match the flat form style
- Fix UI: restore correct square dimensions on the item delete button
- Fix UI: move "Remove All Items" out of the modal header to avoid collision with the close button on mobile

**Screenshots**:
Screenshot 2026-06-27 at 4 04 14 p m
Screenshot 2026-06-27 at 4 04 34 p m
Screenshot 2026-06-27 at 4 04 53 p m


ca-ruz (approved):
LGTM!

Built and tested locally on both mobile and laptop. Works as expected.

Image

Image

client/src/components/pages/Store/Cart/hooks/usePersistentCart.jsx
```diff
@@ -22,6 +23,9 @@ export function usePersistentCart() {
       if (Number.isFinite(storedDiscount)) {
         setDiscount(storedDiscount);
       }
+      if (parsed?.discountType === "fixed" || parsed?.discountType === "percentage") {
```
Vidarte-Alberto: Please use more descriptive name.

Vidarte-Alberto (changes requested):
Please do fixes

Vidarte-Alberto (approved):
LGTM!!

PR #636, opened 2026-07-02
Feature/discount receipt

### LINKED ISSUE #515: Item discounts

Sharmaz:
Changes:
  - Add `discountAmount` field to `TicketData` and render a `DISCOUNT` line above the total in `TOTAL_ROW` when discount is greater than zero
  - Pass `discountAmountCents` through all payment flows (direct, cash, card, BTC) to `printCustomerReceipt` and include it in the ticket data sent to the printer

### DIFF
--- a/client/src/components/pages/Store/Cart/hooks/paymentHandlers.js
+++ b/client/src/components/pages/Store/Cart/hooks/paymentHandlers.js
@@ -138,6 +138,7 @@ export function buildHandlePay({
       await printCustomerReceipt?.({
         items: cartItems,
         totalCents: paymentAmounts.total,
+        discountAmountCents: paymentAmounts.discountAmount,
         ticketId: storeCheckoutResult.ticketId,
       });
 
@@ -205,6 +206,7 @@ async function runDeferredCheckout({
   user,
   printCustomerReceipt,
   refreshShiftTickets,
+  receiptDiscountAmount,
 }) {
   dispatch({ type: "start" });
   try {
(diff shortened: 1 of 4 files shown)


ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):
LGTM!!

PR #638, opened 2026-07-04
Fix: Improve user toast feedback

ca-ruz:
## Summary

  - Add success toast notifications when users are created, updated, or deleted.
  - Keep existing error toast handling in `useUsers`.
  - Rethrow failed create/update/delete requests so modals can react correctly.
  - Keep add, edit, and delete modals open when user mutations fail.
  - Prevent success toasts, form resets, and modal closes on failed user actions.

  ## Files touched

  - `client/src/components/pages/Store/Users/AddUsersModal.jsx`
  - `client/src/components/pages/Store/Users/EditUsersModal.jsx`
  - `client/src/components/pages/Store/Users/Users.jsx`
  - `client/src/components/pages/Store/hooks/useUsers.jsx`
  - Users modal/hook tests

Screenshot From 2026-07-03 18-33-47

Screenshot From 2026-07-03 18-35-03

Screenshot From 2026-07-03 18-35-13

### DIFF
--- a/client/src/components/pages/Store/Users/EditUsersModal.jsx
+++ b/client/src/components/pages/Store/Users/EditUsersModal.jsx
@@ -47,7 +47,11 @@ export function EditUsersModal({ data, setData, roles, onChange, editUsersShowMo
             className="space-y-4"
             onSubmit={async (event) => {
               event.preventDefault();
-              await updateUser(data);
+              try {
+                await updateUser(data);
+              } catch {
+                return;
+              }
               addToast({ description: userTranslations("users.toasts.updateSuccess"), color: "success" });
               setData({
                 userId: "",
--- a/client/src/components/pages/Store/Users/EditUsersModal.jsx
+++ b/client/src/components/pages/Store/Users/EditUsersModal.jsx
@@ -2,14 +2,14 @@
 
 import { useState } from "react";
 
-import { Button, Input, Select, SelectItem, Modal, ModalContent, ModalHeader, ModalBody, ModalFooter } from "@heroui/react";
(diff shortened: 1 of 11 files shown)


Vidarte-Alberto (approved):
LGTM!!

client/src/components/pages/Store/Users/EditUsersModal.jsx
```diff
@@ -55,47 +61,46 @@ export function EditUsersModal({ data, setData, roles, onChange, editUsersShowMo
                 userEmail: "",
                 userRole: "Vendedor",
               });
-              updateUser(data);
               setEditUsersShowModal(false);
             }}
```
Sharmaz: This logic allow the double submit, double http call, double toast.
Also that logic should be in a handler function.

Sharmaz (changes requested):

client/src/components/pages/Store/Users/EditUsersModal.jsx
ca-ruz: Fixed!

ca-ruz:
Update:

  - AddUsersModal.jsx: moved submit logic into handleSubmitAddUser, added immediate ref lock plus loading/disabled state.
  - EditUsersModal.jsx: upgraded the existing submit guard to the same immediate ref lock pattern.
  - DeleteUsersModal.jsx: added guarded handleConfirmDeleteUser, loading/disabled state, and renamed t to userTranslations.
  - AddUsersModal.test.jsx: added double-submit regression test.
  - Users.test.jsx: updated delete modal mock and added double-delete regression test.

Sharmaz (approved):
LGTM

PR #640, opened 2026-07-07
fix(client): share bitcoin price cache and drop stale-price fallback

JordyPirata:
## Summary
- Cart checkout (`useBitcoinInvoice`) and wallet manual invoices (`useWalletAmountInput`/`useSatsToFiatEstimate`) each created their own `BitcoinPriceService` instance with its own 5-minute cache, so the sats amount used to generate an invoice could reflect a BTC price fetched minutes apart between the two flows — the reported "desfase" in price.
- Moved the price cache to module scope (`sharedPriceCache`) so all consumers of `BitcoinPriceService` share the same cached quote, and shrank the cache window from 5 minutes to 1 minute so it refreshes more often.
- Removed the silent fallback to hardcoded `FALLBACK_PRICES` when the CoinGecko fetch fails — `getBitcoinPrice` now always throws on failure instead of quietly returning a stale, hardcoded price. Existing callers already handle the rejected promise (`useBitcoinInvoice`, `useBitcoinPrice`, `useWalletAmountInput`, `useSatsToFiatEstimate`), so this surfaces real failures instead of masking them.
- Added `src/services/__tests__/bitcoinPriceService.test.js` covering: successful fetch, cache reuse within the window, and that fetch/HTTP/missing-currency failures throw instead of returning a fallback price.

## Test plan
- [x] `npm test -- --testPathPatterns="bitcoinPrice|useBitcoinInvoice|useBitcoinPrice|PaymentTab|PaymentConfirmModal|PaymentSentModal|ReceiveTab"` — 113 tests passing
- [x] `npm run lint` — clean

### DIFF
--- a/client/src/services/__tests__/bitcoinPriceService.test.js
+++ b/client/src/services/__tests__/bitcoinPriceService.test.js
@@ -1,0 +1,81 @@
+import BitcoinPriceService from "../bitcoinPriceService";
+
+describe("BitcoinPriceService", () => {
+  let priceService;
+
+  beforeEach(() => {
+    priceService = new BitcoinPriceService();
+    priceService.clearCache();
+    global.fetch = jest.fn();
+    jest.spyOn(console, "warn").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("returns the price from the API on a successful fetch", async () => {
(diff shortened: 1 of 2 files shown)


Vidarte-Alberto (approved):
LGTM!!

Sharmaz (approved):

PR #647, opened 2026-07-10
Fix(Installer): Retry transient phoenixd download failures

ca-ruz:
## Summary

  - Harden phoenixd downloads against transient GitHub/network failures.
  - Increase curl retries from 3 to 5.
  - Add retry delay, retry-all-errors, and connection timeout options.
  - Apply the same retry behavior to both installer paths:
    - `scripts/install.sh`
    - `Makefile install-phoenixd`

  ## Context

  The installer CI failed on macOS ARM64 while downloading the phoenixd release asset from GitHub with a temporary `504` error.
  This change makes the installer more resilient to those transient failures without changing successful install behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,11 @@ install-phoenixd:
 	TMP=$$(mktemp -d); \
 	trap 'rm -rf "$$TMP"' EXIT; \
 	echo "Downloading phoenixd $(PHOENIXD_TAG) ($$ZIP)..."; \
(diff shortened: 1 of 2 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

JordyPirata (approved):
LGTM!!

PR #669, opened 2026-07-22
Feature(scripts): add update.sh with CI validation

JordyPirata:
## Summary

Adds `scripts/update.sh`, a self-contained updater that replaces the currently
installed **phoenixd**, **Ambrosia server** (jar), and **Ambrosia client**
(Next.js build) with the latest GitHub release, following the same conventions
as `scripts/install.sh`. Also consolidates CI so the same workflow that
validates `install.sh` now also validates `update.sh` end-to-end.

### `scripts/update.sh`
- Resolves the latest release tag for both `olympus-btc/ambrosia` and
  `ACINQ/phoenixd` (pinnable via `AMBROSIA_TAG` / `PHOENIXD_TAG`).
- Skips components already at the latest version (override with `--force`).
- Verifies the phoenixd package via the ACINQ PGP key + SHA256SUMS before
  installing.
- Detects installed versions from the jar manifest, the client
  `.ambrosia-version` marker, and `phoenixd --version`.
- Handles systemd services gracefully (stops before update, restarts if they
  were active) and self-heals the client wrapper script/symlink.
- Non-interactive via `--yes`; safe cleanup of temp dirs on exit.

### CI: `Test Install & Update Scripts` (`.github/workflows/test_install.yml`)
Consolidated into the existing install workflow so install + update are covered
together (update depends on a prior install):

- **`lint-scripts`** (ubuntu): `shellcheck` + `bash -n` on `install.sh` and
  `update.sh` — fast, no network.
- **`install-and-update`** (matrix: ubuntu + macOS):
  1. Installs a pinned **older** baseline (`v0.7.0-beta`) via `install.sh`.
  2. Verifies the baseline (symlinks, jar, client deps, exact versions).
  3. Runs `update.sh --yes`.
  4. Verifies server, client, and phoenixd were actually updated to the
     dynamically-resolved latest release, and symlinks remain valid.
  5. **Significance guard**: fails loudly if the pinned baseline ever equals the
     latest release (which would make the update a no-op), prompting a bump of
     `BASELINE_AMBROSIA_TAG`.

## How to test

Locally:
```bash
shellcheck scripts/update.sh
bash -n scripts/update.sh
```

The `install-and-update` job exercises the real path in CI:
`v0.7.0-beta → latest` for server/client and `0.7.2 → latest` for phoenixd on
both Linux and macOS.

## Checklist

- [x] Code follows style guidelines (conventional commits, shellcheck-clean)
- [x] CI validates the change on Linux and macOS


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

PR #674, opened 2026-07-26
Refactor: Share HTTP error handling for failed requests

ca-ruz:
## Summary

  Extracts shared Store HTTP error handling so failed API responses use one consistent Error shape across Store workflows.

  ## Changes

  - Updates `buildHttpError` to return real `Error` objects with `status` and `responseMessage`.
  - Adds `buildParsedHttpError` for failed responses that need backend error messages parsed from JSON.
  - Replaces duplicated local HTTP error builders in Users, Roles, and Refunds.
  - Updates Products to use the new shared helper signature.
  - Adds failed-response handling to Categories mutations so failed create/update/delete responses throw instead of continuing
  as successful.
  - Adds regression coverage for the shared helper and category creation failure behavior.

### DIFF
--- a/client/src/components/pages/Store/Orders/RefundModal.jsx
+++ b/client/src/components/pages/Store/Orders/RefundModal.jsx
@@ -15,18 +15,11 @@ import {
 } from "@heroui/react";
 import { useTranslations } from "next-intl";
 
-import { httpClient, parseJsonResponse } from "@/lib/http";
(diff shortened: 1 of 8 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

PR #685, opened 2026-07-31
feat(client): migrate serwist to turbopack

Vidarte-Alberto:
## Problem

Next.js 16 uses Turbopack by default, but the app was still using the Webpack-based `@serwist/next` integration. Because of that, the production build could stop generating the expected service worker at `/sw.js`, leaving PWA/service-worker behavior unavailable or inconsistent unless the client was built with the temporary `next build --webpack` workaround.

This affected the existing service worker responsibilities, including offline fallback caching and BTC checkout background sync.

## Fix

Migrated the client from `@serwist/next` to `@serwist/turbopack`.

The service worker is now served through a Next route handler at `/serwist/sw.js` instead of being emitted as `public/sw.js`. The app registers it through `SerwistProvider` with root scope (`/`) so it can still control the Store routes and support the existing PWA behavior.

Also added the `Service-Worker-Allowed: /` header to the Serwist route response, which allows the service worker script under `/serwist/sw.js` to register with root scope.

### New files

- **`client/src/app/serwist/[path]/route.js`** - Turbopack-compatible Serwist route handler. It builds the service worker from `src/app/sw.js`, uses `esbuild`, and sets `Service-Worker-Allowed: /`.

### Modified files

- **`client/package.json`** - replaces `@serwist/next` with `@serwist/turbopack`, adds `esbuild`, and aligns `serwist` to `9.5.12`.
- **`client/package-lock.json`** - updates the dependency tree for the Turbopack Serwist integration.
- **`client/next.config.mjs`** - switches from `@serwist/next` to `@serwist/turbopack` and removes the old `swDest: "public/sw.js"` setup.
- **`client/src/app/layout.jsx`** - wraps the app with `SerwistProvider`, registering `/serwist/sw.js` with scope `/` while keeping Serwist disabled in development and Electron runtime.
- **`client/src/app/sw.js`** - updates the worker cache import from `@serwist/next/worker` to `@serwist/turbopack/worker`.
- **`client/eslint.config.mjs`** - ignores generated PWA artifacts so lint remains stable if a generated worker file exists locally.

## How to Test

1. Run:

```bash
cd client
npm run build
```

2. Confirm the build output includes:

```txt
/serwist/[path]
├ /serwist/sw.js.map
└ /serwist/sw.js
```

3. Start the standalone build:

```bash
PORT=3000 node .next/standalone/server.js
```

4. Verify the new service worker route:

```bash
curl -I http://127.0.0.1:3000/serwist/sw.js
```

Expected:

- `HTTP/1.1 200 OK`
- `content-type: application/javascript`
- `service-worker-allowed: /`

5. Verify the old route is no longer used:

```bash
curl -I http://127.0.0.1:3000/sw.js
```

Expected:

- `HTTP/1.1 404 Not Found`

6. Open the app in the browser and check DevTools -> Application -> Service Workers.

Expected:

- script URL is `/serwist/sw.js`
- scope is `/`

7. Confirm existing PWA behavior still works:

- `/offline` still exists
- offline fallback does not show a blank page
- BTC checkout background sync does not show service worker registration errors

8. Run lint:

```bash
cd client
npm run lint
```


client/src/app/serwist/[path]/route.js
```diff
@@ -0,0 +1,23 @@
...
+export const dynamic = "force-static";
+export const dynamicParams = false;
+export const revalidate = false;
+
+export function generateStaticParams() {
+  return serwistRoute.generateStaticParams();
+}
+
+export async function GET(request, context) {
+  const response = await serwistRoute.GET(request, context);
```
Comment (Sharmaz) [inline client/src/app/serwist/[path]/route.js]: Generic name "response" change it.

Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM! 

PR #688, opened 2026-08-03
Refactor: Share CSV download helper

ca-ruz:
## Summary

  Refactors the Reports CSV export flow so Sales and Orders share the same browser download helper.

  ## Changes

  - Adds a shared `downloadCsv` utility for Reports exports.
  - Moves duplicated Blob/object URL/download-link/revoke logic out of:
    - `useSalesData`
    - `useOrdersDetailData`
  - Keeps Sales and Orders CSV content generation inside their existing hooks.
  - Preserves current behavior:
    - successful CSV downloads stay silent
    - failed CSV downloads still use the existing error toast
    - export filenames stay the same
  - Ensures the object URL is revoked with `try/finally` even if the download click fails.
  
   ## Review Follow-up

  - Applied the review suggestion to attach the temporary CSV download anchor to `document.body` before triggering `.click()`.
  - Removed the temporary anchor in the cleanup path after the click attempt.
  - Kept `URL.revokeObjectURL(...)` in `finally` so the Blob URL is released even if the download click fails.
  - Repaired the helper after the GitHub suggested-change application duplicated the function body.
  - Updated the `downloadCsv` unit test to use a real anchor element flow and verify cleanup behavior.

### DIFF
--- a/client/src/components/pages/Store/Reports/utils/downloadCsv.js
+++ b/client/src/components/pages/Store/Reports/utils/downloadCsv.js
@@ -3,21 +3,11 @@ export function downloadCsv(csvContent, filename) {
   const downloadLink = document.createElement("a");
   downloadLink.href = csvDownloadUrl;
   downloadLink.download = filename;
-  try {
-export function downloadCsv(csvContent, filename) {
-  const csvDownloadUrl = URL.createObjectURL(new Blob([csvContent], { type: "text/csv;charset=utf-8;" }));
-  const downloadLink = document.createElement("a");
-  downloadLink.href = csvDownloadUrl;
-  downloadLink.download = filename;
   document.body.appendChild(downloadLink);
   try {
     downloadLink.click();
   } finally {
     downloadLink.remove();
-    URL.revokeObjectURL(csvDownloadUrl);
-  }
-}
-  } finally {
     URL.revokeObjectURL(csvDownloadUrl);
   }
 }
--- a/client/src/components/pages/Store/Reports/utils/downloadCsv.js
+++ b/client/src/components/pages/Store/Reports/utils/downloadCsv.js
@@ -4,7 +4,19 @@ export function downloadCsv(csvContent, filename) {
   downloadLink.href = csvDownloadUrl;
   downloadLink.download = filename;
   try {
+export function downloadCsv(csvContent, filename) {
+  const csvDownloadUrl = URL.createObjectURL(new Blob([csvContent], { type: "text/csv;charset=utf-8;" }));
+  const downloadLink = document.createElement("a");
+  downloadLink.href = csvDownloadUrl;
+  downloadLink.download = filename;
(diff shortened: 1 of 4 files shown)


JordyPirata (approved):
LGTM!!

client/src/components/pages/Store/Reports/utils/downloadCsv.js
```diff
@@ -0,0 +1,11 @@
+export function downloadCsv(csvContent, filename) {
+  const csvDownloadUrl = URL.createObjectURL(new Blob([csvContent], { type: "text/csv;charset=utf-8;" }));
+  const downloadLink = document.createElement("a");
+  downloadLink.href = csvDownloadUrl;
+  downloadLink.download = filename;
+  try {
+    downloadLink.click();
```
Sharmaz, line 8: The anchor is never attached to the document before `.click()` is called. Some browsers (older Firefox in particular) don't reliably fire `click()` on a detached element, so the export can silently do nothing for a subset of users. Since this refactor already centralizes the logic in one place, this is a good spot to fix it too:

  ```suggestion
  export function downloadCsv(csvContent, filename) {
    const csvDownloadUrl = URL.createObjectURL(new Blob([csvContent], { type: "text/csv;charset=utf-8;" }));
    const downloadLink = document.createElement("a");
    downloadLink.href = csvDownloadUrl;
    downloadLink.download = filename;
    document.body.appendChild(downloadLink);
    try {
      downloadLink.click();
    } finally {
      downloadLink.remove();
      URL.revokeObjectURL(csvDownloadUrl);
    }
  }
  ```

Nit: `exportToCsv` in `useSalesData.js` and `useOrdersDetailData.js` still have an identical `catch { addToast({ color: "danger", description: reportsTranslations("export.error") }) }` block after this refactor — the DOM part got deduplicated but the error-handling didn't. Not blocking with only 2 call sites, just flagging in case you want to fold it into `downloadCsv` (e.g. having it accept the toast call, or throw and let a shared wrapper catch it) while you're already touching this code.

Sharmaz (changes requested):

client/src/components/pages/Store/Reports/utils/downloadCsv.js
ca-ruz: Thanks. Applied the anchor attach/remove suggestion.

I’m keeping the toast handling in the hooks for now so `downloadCsv` stays focused on browser download mechanics. The remaining duplicated catch blocks are small and keep translated/user-facing feedback owned by each report hook. (I didn't want to introduce notifications changes in this PR)

ca-ruz:
The GitHub suggestion had accidentally duplicated downloadCsv inside itself, which broke syntax. I cleaned it up and updated the helper test to use a real anchor flow instead of a fake plain object, because the helper now appends the link to document.body.

Sharmaz (approved):
LGTM!!

PR #691, opened 2026-08-04
Feature/wallet backend reconfiguration

Sharmaz:
## Changes:

- Add server-side support to safely rewrite a single property in a conf file, and use it to reconfigure the NWC connection URI without touching other config values
- Propagate translatable error codes when the NWC relay connection fails, and add a way to check whether NWC is the currently active Lightning backend
- Add an authenticated (wallet-password-gated) endpoint to reconfigure the NWC connection URI after initial setup
- Add wallet-service support for the new endpoint and propagate structured error codes from getSeed
- Add a wallet-password-gated NWC Connection card in Settings to update the connection URI, and render it there
- Extract the shared NWC URI regex constant and align WalletBackendStep's styling with the rest of the onboarding wizard
- Stop relying on an unauthenticated getInfo check in the Seed card and show the correct message when the seed is unavailable under NWC

**Screenshots**:

Screenshot 2026-08-03 at 8 57 25 p m

Screenshot 2026-08-03 at 8 58 01 p m

### DIFF
--- a/client/src/components/pages/Onboarding/WalletBackendStep.jsx
+++ b/client/src/components/pages/Onboarding/WalletBackendStep.jsx
@@ -6,7 +6,7 @@ import { Card, CardBody, Input } from "@heroui/react";
 import { Zap, Server } from "lucide-react";
 import { useTranslations } from "next-intl";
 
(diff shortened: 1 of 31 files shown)


JordyPirata (approved):
LGTM!!

ca-ruz (approved):
LGTM! 

PR #698, opened 2026-08-05
Fix/electron skip phoenixd under nwc

Sharmaz:
## Changes:

- Skip starting `phoenixd` in the Electron app when NWC is the configured Lightning backend, instead of always starting it regardless of the user's onboarding choice. The splash screen still reports the step as complete (with an NWC-specific message) so its progress sequence isn't disrupted.
- Skip loading `phoenix.conf` on the server when NWC is configured, avoiding a misleading error log on every boot of a deployment that never has that file.


ca-ruz (approved):
LGTM!

ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):
LGTM!!

PR #699, opened 2026-08-05
Fix/electron preload esbuild bundling

Sharmaz:
## Changes:

- Bundle the preload scripts with esbuild so local `require`s resolve at build time, closing the sandboxed preload bug for good
- Fix macOS dock reactivation after all windows close, and stop Auto Liquidity from silently no-oping when NWC is the configured Lightning backend
- Create the logs directory in `PhoenixdService` before writing to it, matching `BackendService`/`NextJsService`
- Handle 307/308 redirects when downloading the Node.js binary, matching the other download scripts
- Remove the unused `fetchRemoteSha256` helper
- Remove the unused `dev.sh`, `dev.bat`, and `build.sh` scripts


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #700, opened 2026-08-05
Fix/close channel disabled phoenixd

Sharmaz:
## Changes:

- Compare channel state case-insensitively in `ChannelCard.jsx` so the Close Channel button and NORMAL state label work correctly with real phoenixd channels, matching the case-insensitive comparison already used in `NodeSummary.jsx`


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #701, opened 2026-08-06
Chore/admin auth test fixture

Sharmaz:
## Changes:

- Extract the inline `install(Authentication) { ... }` block from `Api().module()` into a top-level `configureAuthentication()` function in `Api.kt`, matching the pattern of its other `configureXxx()` neighbors, so it can be reused by test code without duplicating the auth setup
- Add `AdminAuthTestFixture.kt`, a reusable test helper that seeds a real admin (or non-admin) user via `ExposedTestDb`, generates real `accessToken`/`refreshToken` JWTs via `TokenService`, and installs `configureAuthentication()` on a `testApplication`, enabling end-to-end tests of routes protected by `authenticateAdmin`
- Add `AdminAuthTestFixtureTest.kt` validating the fixture against a minimal `authenticateAdmin`-protected route: unauthenticated requests get 401, non-admin requests get 403, and real admin requests reach the route

### DIFF
--- a/server/app/src/main/kotlin/pos/ambrosia/Api.kt
+++ b/server/app/src/main/kotlin/pos/ambrosia/Api.kt
@@ -63,7 +63,6 @@ public val logger = LoggerFactory.getLogger("Server")
 class Api {
     fun Application.module() {
         AppConfig.loadConfig() // Load the configuration
-        val config = environment.config // Configure the application
         handler() // Install exception handlers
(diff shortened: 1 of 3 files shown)


ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):

Vidarte-Alberto:
LGTM!

PR #703, opened 2026-08-06
fix(client): normalize the NumberInput onChange argument across all call sites

### LINKED ISSUE #689: NumberInput: onChange receives a number from steppers, crashing every event.target handler
**Describe the bug**

Every `NumberInput` in the client that reads `event.target.value` inside an `onChange` handler is
broken, because HeroUI calls that prop from two different places with two different argument types.

From `node_modules/@heroui/number-input/dist/use-number-input.js`:

| Origin | Line | Argument passed to your `onChange` | Fires when |
|---|---|---|---|
| `getInputProps().onChange = chain(inputProps.onChange, onChange)` | 259 | DOM event | typing in the field |
| `useNumberFieldState({ onChange: chain(onValueChange, onChange) })` | 77 | **number** | stepper `+`/`-`, mouse wheel, arrow keys, and the commit on blur/Enter |

So a handler written as `onChange={(event) => … event.target.value}` throws
`TypeError: can't access property "value", event.target is undefined` as soon as the value changes
through the second path.

`onValueChange` is wired only to the numeric path (line 77), which also fires while typing — the
input updates the state and the state notifies. **`onValueChange` alone already covers typing,
steppers, wheel, arrows and commit, so the extra `onChange` is redundant everywhere.**

**To Reproduce**

1. Go to Store → Products.
2. Edit a product that is a bundle pack.
3. Click the `+` / `-` stepper on a component's quantity (or type a value and confirm it).
4. See error: `can't access property "value", event.target is undefined` at
   `BundleProductSelector.jsx:208`.

**Expected behavior**

Changing the quantity with the steppers, the keyboard or the wheel updates the value without
throwing.

**Affected call sites**

*Crash today, no guard:*

- `client/src/components/pages/Store/Products/BundleProductSelector.jsx:208` —
  `onChange={(event) => handleQuantityChange(selectedProduct.productId, Number(event.target.value))}`
  (this is the one reported above; it already has a correct `onValueChange` on line 207)
- `client/src/components/pages/Store/Wallet/Transactions/AmountUnitInputFields.jsx:93` —
  `onChange={(event) => onAmountChange(event.target.value)}`, identical shape, also already has
  `onValueChange`

*Different symptom, same root cause:*

- `client/src/components/pages/Store/Cart/Summary/CartItemCard.jsx:55` — has **only** `onChange` and
  treats the argument as a number: `onChange={(quantityValue) => onUpdateQuantity(cartItem.id, Number(quantityValue))}`.
  It works from the steppers, but while **typing** it receives a DOM event, so `Number(event)` is
  `NaN` and the cart quantity becomes `NaN`.

*Already patched with a defensive `if (event?.target)` guard:*

- `client/src/components/pages/Store/Cart/CashPaymentModal.jsx:100`
- `client/src/components/pages/Store/Orders/CashRefundFields.jsx:25`
- `client/src/components/turn/CloseTurnModal.jsx:121`

These three do not crash, but the guard hides the real cause and the handler still runs twice while
typing (once via `onValueChange`, once via `onChange`).

**Suggested fix**

Drop the `onChange` prop and keep only `onValueChange` at every site. For `CartItemCard`, rename the
handler to `onValueChange` so it keeps receiving a number on both paths. No defensive guards needed.

**Why tests did not catch it**

The 8 test files that mock `NumberInput` all render a plain `<input>`, which only reproduces the DOM
event path — the numeric path is never exercised:

`BundleProductSelector.test.jsx`, `VariantForm.test.jsx`, `CashPaymentModal.test.jsx`,
`CartItemCard.test.jsx`, `SummaryContent.test.jsx`, `CashRefundFields.test.jsx`,
`DiscountInput.test.jsx`, `RefundModal.test.jsx`.

The mock should also expose a stepper that calls `onValueChange(number)` / `onChange(number)`,
mirroring line 77, so this class of bug fails in CI.

**Desktop**

- OS: Linux
- Browser: Firefox
- Version: Next.js 16.2.11 (Turbopack), `@heroui/number-input` as pinned in `client/package.json`

JordyPirata:
## Summary
- Add `toNumberInputValue` to normalize HeroUI's `NumberInput` `onChange`, which hands over a DOM event while typing and a number from the steppers, wheel, arrow keys and the commit on blur/Enter
- Apply it to the 7 affected call sites and drop the defensive `if (event?.target)` guards; `CartItemCard` no longer produces `NaN` quantities and no longer drops the item while the field is empty
- Replace the inline `NumberInput` mocks with a shared one that reproduces the numeric path, so this class of bug fails in CI

Closes #689

## Test plan
- [x] Client tests pass (`npm test`)
- [x] Server tests pass (`./gradlew test`)
- [x] ESLint passes (`npm run lint`)
- [x] ktlint passes (`./gradlew ktlintCheck`)
- [x] Ruff passes (`ruff check .`)
- [x] Build succeeds (`next build` + `./gradlew jar`)


Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #704, opened 2026-08-06
Fix/auto liquidity nwc guard

Sharmaz:
## Changes:

- Skip reading/writing `phoenix.conf` in the `phoenixd:get-auto-liquidity`/`set-auto-liquidity` Electron IPC handlers when NWC is the configured Lightning backend, returning a structured `{ nwcConfigured: true }` signal instead
- Change `useAutoLiquidity` to load its state on demand via a new `load()` function instead of on mount, so loading can be deferred until wallet access is confirmed; both `load()` and `toggle()` now recognize the NWC-unavailable signal from the backend
- Add `LightningCardLocked`/`LightningCardUnlocked`, gating Auto Liquidity behind a wallet-password prompt (`WalletGuard`), matching the existing pattern used by the Seed and NWC Connection cards
- Rewire `LightningCard` as an orchestrator that starts locked and only loads/exposes the Auto Liquidity toggle after wallet authorization, showing a toast and returning to the locked state if NWC is configured
- Update `Settings` tests to match `LightningCard`'s new locked-by-default rendering

**Screenshots**:

imagen


client/src/hooks/useAutoLiquidity.js
```diff
@@ -43,6 +51,9 @@ export function useAutoLiquidity() {
     try {
       const value = newEnabled ? "2m" : "off";
```
Vidarte-Alberto: Please use more descriptive names.

Vidarte-Alberto (changes requested):
Do fixes

Vidarte-Alberto (approved):
LGTM!

ca-ruz (approved):
LGTM!

PR #705, opened 2026-08-06
Fix/wallet node unavailable hang

Sharmaz:
## Changes:

- Fix `StoreWallet.jsx` so the `/store/wallet` page no longer hangs on the "Loading wallet info..." spinner when the active Lightning backend (NWC or phoenixd) fails to respond — a new `infoLoading` flag (cleared in `fetchInfo`'s `finally`) replaces the old `if (!info)` render gate, which never released on error and left `NodeError` permanently unreachable
- Parallelize `fetchTransactions`'s incoming/outgoing fetches with `Promise.all` instead of awaiting them sequentially, halving the worst-case wait under NWC's per-request timeout
- Generalize the `nodeInfo.nodeUnavailable` message (`NodeInfo/locales/en.js`/`es.js`) to no longer assume phoenixd specifically, since the error card is now reachable for both backends
- Update `StoreWallet.test.jsx`: strengthen the error-handling tests to assert the actual resulting UI (`NodeError` text, spinner gone, transactions empty-state fallback) instead of only checking that a mock was called, and add a regression test covering the new concurrent transaction fetch

**Screenshots**:

imagen

### DIFF
--- a/client/src/components/pages/Store/Wallet/NodeInfo/locales/en.js
+++ b/client/src/components/pages/Store/Wallet/NodeInfo/locales/en.js
@@ -14,7 +14,7 @@ const nodeInfoEn = {
     inboundLiquidity: "Inbound Liquidity:",
     fetchInfoError: "Error fetching wallet information",
     getInfoErrorDescription: "Could not load wallet information",
-    nodeUnavailable: "Lightning node is not available. Make sure phoenixd is running.",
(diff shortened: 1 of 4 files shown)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #707, opened 2026-08-07
Fix: Update active category filter state in cart

ca-ruz:
## Summary

  Fixes the cart/sales product category filter so the active pill reflects the selected category instead of always showing “All” as active.

  ## Changes

  - Updates the “All” filter button to show active styling only when no category filter is selected.
  - Updates category filter buttons to show active styling when their category is selected.
  - Adds `aria-pressed` to the filter buttons so the selected state is reflected in the DOM.
  - Cleans up touched callback variable names in `SearchProducts.jsx`.

  ## Testing

  - Added regression coverage for the category filter selected state.
  - Verified that selecting a category moves the active state from “All” to that category.
  - Verified that selecting “All” restores the active state to “All”.

Screenshot From 2026-08-07 09-38-45

Screenshot From 2026-08-07 09-39-01

Screenshot From 2026-08-07 09-39-07

CI note: the Electron npm audit failure is unrelated to this PR. This branch only changes the Store cart category filter UI snd its focused test. The audit failure comes from existing Electron/js-yaml dependency advisories and should be handled in a separate dependency PR.

### DIFF
--- a/client/src/components/pages/Store/Cart/SearchProducts.jsx
+++ b/client/src/components/pages/Store/Cart/SearchProducts.jsx
@@ -15,8 +15,8 @@ export function SearchProducts({ products, onAddProduct, categories }) {
   const filteredProducts = products.filter((product) => {
     const categoryIds = product.categoryIds ?? [];
     const categoryNames = categories
-      .filter((cat) => categoryIds.includes(cat.id))
-      .map((cat) => cat.name)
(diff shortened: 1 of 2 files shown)


Sharmaz (approved):
LGTM!!

Vidarte-Alberto (approved):
LGTM!!

PR #709, opened 2026-08-07
fix(client): move serwist build deps to production

Vidarte-Alberto:
This pull request updates the dependency management in `client/package-lock.json` by adjusting how certain packages are included and removing the `"dev": true` flag from many dependencies. It also reorders where some dependencies are listed, which may affect how they are installed or bundled.

Dependency updates and management:

* Moved `@serwist/turbopack` and `esbuild` from the `devDependencies` section to the main dependencies section, making them regular dependencies instead of development-only.
* Removed the `"dev": true` flag from a large number of dependency entries, which changes their classification from development-only to production dependencies. This affects packages such as `@serwist/turbopack`, `@serwist/build`, `@serwist/window`, `browserslist`, `glob`, `lru-cache`, `path-scurry`, `semver`, `source-map`, `type-fest`, `zod`, `@types/trusted-types`, `balanced-match`, `brace-expansion`, and many architecture-specific sub-dependencies. 
These changes will result in certain packages being installed in production environments rather than only during development, which may impact build times and deployed bundle size. Be sure to verify that all dependencies are needed in production before merging.


Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #711, opened 2026-08-11
Fix(client): user auth admin escalation

Vidarte-Alberto:
## Problem

The user-management API exposed two critical privilege and session-token risks.

First, `GET /users/{id}` was publicly accessible and returned the user's stored `refreshToken`. An unauthenticated attacker could call the intentionally public `GET /users` endpoint to discover user IDs, request an administrator by ID, obtain the administrator refresh token, and exchange it for an authenticated session.

Second, permissions such as `users_update` and `roles_update` were enough to grant administrator privileges. A non-admin user with one of those permissions could assign an existing admin role to a user or promote a regular role by setting `isAdmin: true`. The equivalent create operations had the same problem: `users_create` could create a user with an admin role, and `roles_create` could create a new admin role.

The client also showed only a generic error when the server rejected one of these operations, giving the user no explanation that administrator privileges were required.

## Fix

Protected individual user reads and prevented non-admin users from creating or assigning administrator privileges.

`GET /users/{id}` now requires the `users_read` permission, while the intentionally public `GET /users` endpoint remains unchanged for the pre-login user/PIN selector. `getUserById()` now masks both `pin` and `refreshToken`, matching the defense-in-depth behavior already used by the public user list.

Admin role creation and assignment now require the authenticated caller to already be an administrator. This applies to both user and role creation/update operations. The check uses the existing server-side admin guard, which resolves the caller from the stored refresh token and current database role instead of trusting only the access-token claim.

The client now shows localized warning toasts when user or role management receives the admin-only `403` response.

### Protected individual user lookup

Moved `GET /users/{id}` behind `authorizePermission("users_read")`.

The public `GET /users` route remains available because it is required before login. Its existing masking behavior was not changed.

### Refresh token masking

Updated `UsersService.getUserById()` so it returns:

- `pin: "****"`
- `refreshToken: "****"`

Raw session tokens are no longer exposed by the user response, even to an authenticated caller.

### Admin-only user role assignment

Added an admin check when:

- creating a user with an admin role through `POST /users`
- assigning an admin role through `PUT /users/{id}`

Users with `users_create` or `users_update` can continue managing ordinary users and roles, but they cannot grant administrator access unless they are already administrators.

### Admin-only role creation and promotion

Added an admin check when:

- creating a role with `isAdmin: true` through `POST /roles`
- promoting a role with `isAdmin: true` through `PUT /roles/{id}`

Users with `roles_create` or `roles_update` can continue managing standard roles without being able to escalate privileges.

### Localized client feedback

Added specific English and Spanish warning toasts for rejected administrator assignments.

The user-management flow now explains that only an administrator can assign an admin role. The role-management flow explains that only an administrator can create or grant an admin role. Existing generic, duplicate-name, and last-admin feedback remains unchanged.

### Security regression coverage

Added HTTP route tests covering:

- the public user list remaining accessible without authentication
- individual user lookup rejecting unauthenticated requests
- `users_create` not allowing a non-admin to create an admin user
- `users_update` not allowing a non-admin to assign an admin role
- `roles_create` not allowing a non-admin to create an admin role
- `roles_update` not allowing a non-admin to promote a role

Added service coverage confirming that `getUserById()` masks both the PIN and refresh token.

Added client coverage for the administrator-required toasts in user creation/update and role creation/promotion flows.

## New Files

- `server/app/src/test/kotlin/pos/ambrosia/utest/UserPrivilegeEscalationRouteTest.kt` — HTTP regression coverage for authentication and privilege-escalation attempts.

## Modified Files

- `server/app/src/main/kotlin/pos/ambrosia/api/Users.kt` — protects user-by-ID reads and requires admin status when creating or assigning admin users.
- `server/app/src/main/kotlin/pos/ambrosia/api/Roles.kt` — requires admin status when creating or promoting admin roles.
- `server/app/src/main/kotlin/pos/ambrosia/services/UsersService.kt` — masks refresh tokens and exposes the role-admin lookup used by the route guard.
- `server/app/src/test/kotlin/pos/ambrosia/utest/UsersServiceTest.kt` — verifies sensitive user fields are masked.
- `client/src/components/pages/Store/hooks/useUsers.jsx` — shows administrator-required feedback for rejected user mutations.
- `client/src/components/pages/Store/hooks/__tests__/useUsers.test.jsx` — verifies the user-management warning toast.
- `client/src/components/pages/Store/Users/Roles/Roles.jsx` — shows administrator-required feedback for rejected role mutations.
- `client/src/components/pages/Store/Users/Roles/__tests__/Roles.test.jsx` — verifies role creation and promotion warning toasts.
- User and role locale files — add English and Spanish administrator-required messages.

## Screenshots
image

## How to Test

1. Create a standard role with only `users_read`, `users_create`, `users_update`, `roles_read`, `roles_create`, and `roles_update` permissions.

2. Assign that role to a non-admin user and log in as that user.

3. Confirm the public login user selector still loads before authentication.

4. While logged in, confirm users with `users_read` can access individual user information and that no raw PIN or refresh token is returned.

5. Try to create a user with an existing admin role.

6. Confirm the request is rejected with `403 Forbidden` and the client shows the administrator-required warning.

7. Try to assign an existing admin role to a regular user.

8. Confirm the request is rejected with `403 Forbidden` and the client shows the administrator-required warning.

9. Try to create a role with **With admin privileges** enabled.

10. Confirm the request is rejected with `403 Forbidden` and the client shows the administrator-required warning.

11. Try to edit a standard role and enable **With admin privileges**.

12. Confirm the request is rejected with `403 Forbidden` and the client shows the administrator-required warning.

13. Log in as an administrator and confirm all four administrator creation/assignment operations still succeed.

14. Confirm non-admin user and role operations continue to work when the caller has the corresponding permission.

## Verification Run

```bash
cd server
./gradlew test ktlintCheck

cd ../client
npm test
npm run lint

cd ..
git diff --check
```

Verification results:

- server test suite passed
- server `ktlintCheck` passed
- 264 client test suites passed
- 2,515 client tests passed
- client ESLint passed
- `git diff --check` passed

## Notes

- `GET /users` remains public by design because the PIN login screen uses it before authentication. It continues masking `pin` and `refreshToken`.
- This change is separate from hashing refresh tokens at rest. Masking the HTTP response prevents direct token disclosure, while hashing stored refresh tokens remains a separate defense for database compromise.
- Existing last-admin protections remain in place and continue preventing deletion or demotion of the final active administrator.
- This PR does not add step-up authentication. It enforces that the caller is already an administrator before administrator privileges can be granted.

### DIFF
--- a/client/src/components/pages/Store/Users/Roles/Roles.jsx
+++ b/client/src/components/pages/Store/Users/Roles/Roles.jsx
@@ -73,10 +73,15 @@ export function Roles({ roles, createRole, deleteRole, loading: loadingRoles, up
       setShowModal(false);
       addToast({ title: roleTranslations("roles.actions.createSuccess"), color: "success" });
     } catch (error) {
+      const adminRequired = error?.status === 403 && form.isAdmin;
       addToast({
-        title: error?.status === 409 ? roleTranslations("roles.actions.createConflictTitle") : roleTranslations("roles.actions.createErrorTitle"),
-        description: error?.status === 409 ? roleTranslations("roles.actions.createConflictDescription") : roleTranslations("roles.actions.createErrorDescription"),
-        color: error?.status === 409 ? "warning" : "danger",
+        title: adminRequired
+          ? roleTranslations("roles.actions.adminRequiredTitle")
+          : error?.status === 409 ? roleTranslations("roles.actions.createConflictTitle") : roleTranslations("roles.actions.createErrorTitle"),
+        description: adminRequired
+          ? roleTranslations("roles.actions.adminRequiredDescription")
+          : error?.status === 409 ? roleTranslations("roles.actions.createConflictDescription") : roleTranslations("roles.actions.createErrorDescription"),
+        color: adminRequired || error?.status === 409 ? "warning" : "danger",
       });
     } finally {
       creatingRef.current = false;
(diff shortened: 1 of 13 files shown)


Sharmaz (approved):
LGTM!!

JordyPirata (approved):
LGTM!!

PR #717, opened 2026-08-14
feat: enforce 6-digit PIN across server, client and e2e tests

JordyPirata:
## Description

Migrates the user PIN from 4 to 6 digits across the whole stack.

- **server**: `Users.kt` / `UsersService.kt` validate that the PIN is exactly 6 numeric digits.
- **client (onboarding and user management)**: `Onboarding`, `AddUserAccount`, `AddUsersModal` and `EditUsersModal` require 6 digits; `en`/`es` locale strings updated.
- **client (login)**: `PinPad` accepts 6 digits, plus a new `PinDeprecationModal` that warns users holding a legacy 4-digit PIN when they sign in.
- **tests**: client unit tests and e2e fixtures (`conftest.py`, `auth_utils.py`, `test_pin_validation_e2e.py`, etc.) updated to 6-digit PINs.

## Backwards compatibility with legacy 4-digit PINs

`POST /auth/login` deliberately does **not** validate length: users created before this change need to sign in precisely so they can update their PIN — that sign-in is what triggers the `PinDeprecationModal`. The 6-digit rule applies only when creating or updating a user; `AuthService` and `SecurePinProcessor` stay length agnostic.

That contract previously held only by omission, with no test protecting it. It is now pinned down by:

- `AuthLoginRouteTest` — `login accepts a legacy four digit pin` covers the route itself, which is where a length check would realistically be added.
- `UserPinValidationRouteTest` — the full migration path: a legacy user signs in with 4 digits, does `PUT /users/{id}` with 6, and the old PIN stops working from then on.
- `AuthServiceTest` — the same guarantee at the service layer.

Verified by mutation testing: adding `if (pin.length != 6)` to `Authorize.kt` kills exactly the legacy PIN test while the 6-digit one still passes.

## Images
image

## Server unit tests

The Kotlin tests had been left behind with 4-digit literals. `UserPrivilegeEscalationRouteTest` was failing: the PIN check in `Users.kt` runs **before** `requireAdmin()`, so the request died with a 400 and never reached the 403 the test asserts.

- Literals updated in `UserPrivilegeEscalationRouteTest` and `TimezoneValidationRouteTest`.
- New coverage of the 6-digit rule in `UsersService.addUser` and in the `POST`/`PUT /users` routes.
- Added `ExposedTestDb.seedUserWithPin`: `seedUser` stored the literal `"****"` in the `pin` column, so no seeded user could ever authenticate and there was no hash/verify round-trip coverage at any length.

Server suite: **739 tests, 0 failures** (previously 723 with 1 failure). `ktlintCheck` clean.

## Login logging

`Authorize.kt` dumped the whole `AuthResponse` on every successful login — that object carries `email` and `phone`, putting PII into the logs in clear text. It now logs the user id only, per the "log identifiers.
## Status

Branch is up to date with `development` (0 commits behind, no conflicts). The conflicts in `PinLogin.jsx`, `AddUsersModal.jsx` and `EditUsersModal.jsx` that were blocking it earlier are resolved.

Pending: take it out of draft once full QA passes.

## How to test

1. Onboarding: creating a user with a 4-digit PIN must be rejected; 6 digits must succeed.
2. Sign in as a user holding a legacy 4-digit PIN → login succeeds and the deprecation modal appears.
3. Store → Users: both create and edit require 6 digits.
4. `cd server && ./gradlew test`
5. `cd server/e2e_tests_py && pytest tests/test_pin_validation_e2e.py`


server/app/src/test/kotlin/pos/ambrosia/utest/AuthServiceTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    private val testEnv =
+        applicationEnvironment {
+            config = MapApplicationConfig("secret" to "auth-service-test-secret")
+        }
+    private val service = AuthService(testEnv)
+    private lateinit var dbFile: File
+
+    @Before
+    fun setUp() {
+        dbFile = ExposedTestDb.connect()
```
Sharmaz: Use `databaseFile` instead of `dbFile` as in `server/app/src/test/kotlin/pos/ambrosia/utest/UserPinValidationRouteTest.kt`

server/e2e_tests_py/ambrosia/auth_utils.py
```diff
@@ -10,7 +10,7 @@
 from ambrosia.http_client import AmbrosiaHttpClient
 
 # Default test user credentials
```
Sharmaz: DEFAULT_TEST_USER is enough clear to rid of the comment line.

Sharmaz (changes requested):

Sharmaz (approved):
LGTM!!

ca-ruz (commented):
ACK.

Image

Image

Image

PR #721, opened 2026-08-15
Fix/store permission crash

Sharmaz:
## Changes:

- Let individual fetches opt out of the global 403 redirect, the foundation the rest of this work builds on
- Stop the Store dashboard from crashing when a card's underlying permission is missing
- Stop Products from crashing when the user lacks categories_read
- Stop Roles from crashing when the user lacks permissions_read
- Auto-suggest cart-related permissions (products_read, categories_read, payments_read) when granting orders_create in the role editor
- Stop Orders and Cart from crashing when the user lacks payments_read
- Expose whether useCategories was blocked by a missing permission
- Stop the payment currency lookup from crashing Cart when payments_read is missing
- Show an explicit message in Cart when a permission is missing instead of crashing
- Require users_read for the users list endpoint, on both server and client
- Require printer_update for printer and ticket template mutations, without blocking routine printing, on both server and client
- Add settings_read to the permission catalog so it's grantable to non-admin roles, and clean up naming in Currency.kt
- Use meaningful names for translation, element, and event handler variables in the ticket template editor
- Use meaningful names in Cart, Store hooks, and useFetchList test mocks found during review
- Use a grantable baseline permission in the users_read e2e tests instead of the now admin-only permissions_read
- Add a public minimal users endpoint for the PIN login screen, replacing the now permission-gated users list it depended on, on both server and client
- Stop the open-shift check from redirecting to /unauthorized when shifts_read is missing
- Stop WalletGuard's automatic logout from redirecting to /unauthorized

### DIFF
--- a/client/src/components/pages/Auth/PinLogin/PinLogin.jsx
+++ b/client/src/components/pages/Auth/PinLogin/PinLogin.jsx
@@ -9,7 +9,7 @@ import { useTranslations } from "next-intl";
 
 import { useAuth } from "@/hooks/auth/useAuth";
 import { useConfigurations } from "@/providers/configurations/configurationsProvider";
-import { getUsers } from "@/services/authService";
+import { getPublicUsers } from "@/services/authService";
 
 import { BusinessHeader } from "./BusinessHeader";
 import { EmployeeSelect } from "./EmployeeSelect";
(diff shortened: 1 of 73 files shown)


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

Built and tested locally, everything works as before, it seems nothing broke.

PR #724, opened 2026-08-18
Fix/fetch list forbidden handling

Sharmaz:
## Changes:

- Make `useFetchList` throw on failed responses instead of silently returning `null` (with matching toast copy), and fix `useOrders`' dead error-handling branch that depended on the old behavior.
- Remove dead code found while auditing `useFetchList`'s consumers: the unused `usePayments` hook, and unread `error` state in `useBitcoinInvoice` and `useRoles`.
- Sweep the remaining `useFetchList` consumers (`usePaymentMethod`, `usePrinter`, `useTemplates`, `useCategories`) for the same dead null-check and missing test coverage.
- Stop Users' and Roles' mutations from redirecting to `/unauthorized` on a 403, and add a shared `PermissionBlockedMessage` component shown on the Users page instead.
- Stop Products', Categories', and Product Variants' mutations from redirecting to `/unauthorized` on a 403, showing the same blocked message on the Products page.
- Stop Orders' refund requests and its list read from redirecting to `/unauthorized` on a 403, showing the same blocked message on the Orders page.
- Stop background and mutation requests across Settings (printers, templates, config), Cart (payment currency, BTC checkout recovery), Wallet login, and Reports from redirecting to `/unauthorized` on a 403.

**Screenshots**:

imagen


ca-ruz (approved):
LGTM!

Built and tested locally, everything works as before, it seems nothing broke.

Vidarte-Alberto (approved):
LGTM!!

PR #725, opened 2026-08-18
chore(ci): pin actions to SHA, add CodeQL and secret scanning, ratchet coverage

### LINKED ISSUE #723: E2E: server startup times out on a cold build and leaks an orphan JVM
**Describe the bug**

Running the Python E2E suite locally fails in two related ways, both rooted in
`AmbrosiaTestServer` (`server/e2e_tests_py/ambrosia/test_server.py`).

**1. `STARTUP_TIMEOUT` is too short for a cold start**

`STARTUP_TIMEOUT = 30` (line 40) while the server is started with
`./gradlew run --no-daemon` (lines 68-74). On a cold tree, Gradle has to configure and compile the
Kotlin sources before Ktor ever binds the port, and that does not fit in 30 s — the whole suite
errors out at collection time:

```
242 items collected
...
3 skipped, 239 errors in 37.63s
RuntimeError: Server did not start within 30 seconds
```

The CI workflow hides this because it runs `./gradlew build -x test` in a separate step before
`pytest` (`.github/workflows/e2e.yml`), so by the time the fixture starts the server the compile is
already done. Nothing in the test setup or in `server/e2e_tests_py/CLAUDE.md` states that
pre-building is required — the documented flow is just `source .venv/bin/activate && pytest`.

Pre-building fixes the session-scoped server, but not the second one: `test_nwc_backend_e2e.py`
boots an independent instance on port 9165 via the same `./gradlew run --no-daemon` **while the main
server is still running**. That contention pushes it past 30 s too, so with a warm build the suite
still ends as:

```
237 passed, 3 skipped, 2 errors in 98.96s
ERROR tests/test_nwc_backend_e2e.py::TestNwcBackend::test_getinfo_fails_when_the_relay_is_unreachable - RuntimeError: Server did not start within 30 seconds
ERROR tests/test_nwc_backend_e2e.py::TestNwcBackend::test_close_channel_returns_unsupported_operation - RuntimeError: Server did not start within 30 seconds
```

Both NWC tests pass when the file is run on its own (`2 passed in 27.25s`) — 27 s, i.e. just under
the limit even without contention. The margin is thin enough that these two are effectively flaky
on any machine slower than the one that set the number.

**2. A timed-out server is leaked as an orphan process**

When `_wait_for_server` gives up it raises at line 161. `start_server` catches that and calls
`self._cleanup_server()` (lines 95-98), but `_cleanup_server` only clears the Python reference:

```python
def _cleanup_server(self) -> None:
    """Clean up server process references."""
    if self.server_process:
        self.server_process = None
```

The `subprocess.Popen` process group is never signalled — unlike the `stop_server` path, which does
`os.killpg(...)`. So the JVM keeps booting, binds the port a few seconds later, and stays alive
after pytest exits. After the two runs above I had two orphans holding 9154 and 9165/9455:

```
$ ss -ltnp | grep -E "9154|9165|9455"
LISTEN 0 4096 [::ffff:127.0.0.1]:9455 *:* users:(("java",pid=200073,fd=213))
LISTEN 0 4096 [::ffff:127.0.0.1]:9165 *:* users:(("java",pid=200073,fd=212))
```

Both had to be killed by hand. This also makes the next run misleading rather than merely failing:
an orphan on 9154 answers `200` on the health check, so a subsequent session would happily run the
whole suite against a stale server it does not own, with a datadir it never initialized.

**To Reproduce**

1. Start from a tree with no compiled server output (or run `./gradlew clean` in `server/`).
2. `cd server/e2e_tests_py && source .venv/bin/activate && pytest tests/`
3. Every test errors with `RuntimeError: Server did not start within 30 seconds`.
4. `ss -ltnp | grep 9154` — a `java` process is still listening after pytest exited.

**Expected behavior**

- The suite starts the server successfully from a cold tree, without an undocumented manual
  pre-build step.
- A startup timeout leaves no orphan JVM behind.

**Suggested fix**

- Raise `STARTUP_TIMEOUT` to something that covers a cold Gradle configure + compile (120 s or
  more), or split the wait into "process compiled" and "port answering" phases so a slow compile
  does not read as a dead server.
- Make the failure path in `start_server` terminate the process group the way `stop_server` does,
  instead of only dropping the reference — e.g. call `_force_kill_server()` before
  `_cleanup_server()` in the `except` block.
- Optionally, have `_wait_for_server` fail fast with a clear message when the target port is already
  held by a process the fixture did not spawn, so a leaked orphan cannot be silently reused.

**Environment**

- Linux, Java 21 (temurin 21.0.8), Python 3.12.3, pytest 9.0.3
- Observed on branch `feat/6-digit-pin`; unrelated to that branch's changes — the failure is in the
  test harness, and all PIN-related E2E tests (`test_pin_validation_e2e.py`) pass once the server
  starts.

JordyPirata:
## What this does

Hardens CI, adds the two security scans we were missing, and puts a coverage ratchet on both client and server.

Ambrosia custodies a self-custodial Lightning node, which makes the CI supply chain a real attack surface. An unpinned `actions/checkout@v5` runs whatever that tag points at the moment the job starts.

### GitHub Actions supply chain

- Every action pinned to a full SHA with the tag in a trailing comment (`actions/checkout@93cb6ef… # v5.0.1`, etc.) across `client_linter`, `client_unit_test`, `e2e`, `server_linter`, `server_unit_tests`, `test_install`, `test_install_from_source`.
- Explicit workflow-level `permissions: contents: read` where it was missing, so `GITHUB_TOKEN` no longer inherits default write scopes.
- New `.github/dependabot.yml`: npm (`/client`, `/electron`), gradle (`/server`), uv (`/server/e2e_tests_py`) and github-actions, all with `target-branch: development` and conventional commit prefixes per scope. Minor and patch updates are grouped so we don't drown in PRs.

### New scans

- **`codeql.yml`** — `security-and-quality` analysis over `java-kotlin` (manual build via `./gradlew :app:jar`) and `javascript-typescript`. Runs on PRs, on push to `main`/`development`, and weekly — the scheduled scan is what catches newly published queries against code that hasn't changed.
- **`secret_scan.yml`** + **`.gitleaks.toml`** — gitleaks 8.30.1. On a PR it scans only the commits the PR introduces (`origin/$base..HEAD`) rather than full history, so the job fails on *new* findings instead of pre-existing ones. The config extends the default ruleset and only adds *path* allowlists: build artifacts, and the NWC test vectors (fixed hex keypairs driving a handshake against a local relay, with no funds behind them).

### Coverage ratchet

Thresholds sit **below** the measured baseline — they're a regression guard, not a target, so real regressions fail while normal churn doesn't.

| | baseline | threshold |
|---|---|---|
| client statements / branches / functions / lines | 88.36 / 79.35 / 88.36 / 90.11 | 85 / 76 / 85 / 87 |
| server line / branch | 62.5 / 26.9 | 58 / 23 |

- `client/jest.config.mjs`: adds `coverageThreshold`; the workflow moves from `npm test` to `npm run test:coverage`.
- `server`: Kover plugin (`0.9.9`) with `koverVerify` rules, a new step in `server_unit_tests.yml`, and the report uploaded as an artifact (7-day retention).

### Also

- `dependency_review.yml` now runs on PRs against any branch (`'**'`), not just `main`/`development`.

## Why the config changes ship together

`client/jest.config.mjs`, `server/app/build.gradle.kts` and `server/gradle/libs.versions.toml` aren't workflow changes, but the new CI steps (`npm run test:coverage`, `:app:koverVerify`) don't exist without them. Splitting them would leave CI broken in the intermediate PR.

## Verification

- All 12 workflows and `dependabot.yml` parse as valid YAML.
- Every action in the repo is SHA-pinned (including `electron_linter.yml` and `release.yml`, which already were); no workflow is left without a `permissions` block.
- `npm run test:coverage` green (274 suites, 2586 tests) and `./gradlew :app:koverVerify` BUILD SUCCESSFUL — the new thresholds pass against current code.

## Checklist

- [x] Code follows style guidelines
- [x] Tests pass locally
- [x] Documentation is updated if needed (inline comments in each config explain why each threshold sits where it does)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,67 @@
+version: 2
+
+updates:
+  - package-ecosystem: npm
+    directory: /client
+    target-branch: development
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(client)"
+    groups:
+      client-minor-and-patch:
+        update-types:
+          - minor
+          - patch
+
+  - package-ecosystem: npm
+    directory: /electron
+    target-branch: development
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(electron)"
+    groups:
+      electron-minor-and-patch:
+        update-types:
+          - minor
+          - patch
+
+  - package-ecosystem: gradle
+    directory: /server
+    target-branch: development
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 5
+    commit-message:
+      prefix: "chore(server)"
+    groups:
+      server-minor-and-patch:
+        update-types:
+          - minor
+          - patch
+
+  # e2e_tests_py uses uv (pyproject.toml + uv.lock), not pip/requirements.txt.
+  - package-ecosystem: uv
+    directory: /server/e2e_tests_py
+    target-branch: development
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 3
+    commit-message:
+      prefix: "chore(e2e)"
+
+  - package-ecosystem: github-actions
+    directory: /
+    target-branch: development
+    schedule:
+      interval: weekly
+      day: monday
+    open-pull-requests-limit: 3
+    commit-message:
+      prefix: "chore(ci)"
--- a/.github/workflows/client_linter.yml
+++ b/.github/workflows/client_linter.yml
@@ -5,13 +5,16 @@ on:
     branches:
       - '**'
 
+permissions:
+  contents: read
+
 jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
 
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           node-version: '24'
           cache: 'npm'
--- a/.github/workflows/client_unit_test.yml
+++ b/.github/workflows/client_unit_test.yml
@@ -7,13 +7,16 @@ on:
 
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
 
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
         with:
           node-version: '24'
           cache: 'npm'
@@ -25,5 +28,6 @@ jobs:
       - run: npm audit --omit=dev --audit-level=high
         working-directory: ./client
 
-      - run: npm test
+      # Coverage thresholds live in client/jest.config.mjs and fail the run on regression.
+      - run: npm run test:coverage
         working-directory: ./client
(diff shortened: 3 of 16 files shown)


Sharmaz (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #727, opened 2026-08-18
Fix/background forbidden redirect

Sharmaz:
## Changes:

- Stop `ticketsService`'s background requests (tickets, payments, payment methods, orders) from forcing a redirect to `/unauthorized` on a 403 — these run globally via `TurnProvider` whenever a shift is open, so a missing permission previously broke navigation on every `/store/*` page, not just one.
- Stop `adminNotificationsService`'s requests from forcing the same unwanted redirect on a 403.
- Add error handling to `VariantManager`'s `executeVariantMutation`, which previously let unexpected mutation failures (e.g. network errors) fail silently; it now shows an error toast and keeps the edit form open instead of appearing to succeed.


Vidarte-Alberto (approved):
LGTM

ca-ruz (approved):
LGTM!

PR #728, opened 2026-08-19
Fix/settings read navigation

Sharmaz:
## Changes:

- Gate `/store/settings` navigation by `settings_read` instead of `settings_update`, fixing a phantom-permission bug where a role granted only view access to Settings couldn't reach the page at all.
- Require `settings_update` to edit StoreInfo in Settings; the read-only gate above no longer implies write access.
- Require `settings_update` to edit Currency (keeps the current value visible but disables the selector), and stop Currency's requests from redirecting to `/unauthorized` on a 403.
- Require `settings_update` to edit NwcConnection, and stop NwcConnection/Seed requests from redirecting to `/unauthorized` on a 403.
- Stop Printers and TicketTemplates list requests from redirecting to `/unauthorized` on a 403 (their write actions were already gated).
- Hide Seed, NwcConnection, and Tutorials from non-admin roles in Settings, since the wallet operations they surface require admin regardless of any RBAC permission.

### DIFF
--- a/client/src/components/pages/Store/Settings/StoreInfo/StoreInfoCard.jsx
+++ b/client/src/components/pages/Store/Settings/StoreInfo/StoreInfoCard.jsx
@@ -5,9 +5,8 @@ import Image from "next/image";
 import { Button, Card, CardBody, CardFooter, CardHeader } from "@heroui/react";
 import { useTranslations } from "next-intl";
 
+import { storedAssetUrl } from "@/components/utils/storedAssetUrl";
 import { RequirePermission } from "@/hooks/usePermission";
-
-import { storedAssetUrl } from "../../../../utils/storedAssetUrl";
 
 export function StoreInfoCard({ data, onEdit }) {
   const settingsTranslations = useTranslations("settings");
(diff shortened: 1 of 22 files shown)


client/src/components/pages/Store/Settings/StoreInfo/StoreInfoCard.jsx
```diff
@@ -5,6 +5,8 @@ import Image from "next/image";
 import { Button, Card, CardBody, CardFooter, CardHeader } from "@heroui/react";
 import { useTranslations } from "next-intl";
 
+import { RequirePermission } from "@/hooks/usePermission";
+
 import { storedAssetUrl } from "../../../../utils/storedAssetUrl";
```
Vidarte-Alberto: Please use aliases "@" on imports

Vidarte-Alberto (changes requested):

Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #729, opened 2026-08-19
Fix/wallet forbidden access

Sharmaz:
## Changes:
- Require `wallet_read` to reach `/store/wallet`; a role without it now sees a permission-blocked message instead of the wallet password prompt, matching the same pattern already used on Users/Products/Orders.

**Screenshots**:

imagen


Vidarte-Alberto (approved):
LGTM!!

ca-ruz (approved):
LGTM!

PR #736, opened 2026-08-27
chore(server): route coverage

JordyPirata:
## Why

We want to make architecture changes to the server in the medium term, and that needs a safety net
first. The problem was never the total coverage but its distribution: the service layer sat at 88%
while the HTTP layer sat at **12.3%** — the wrong way round for a refactor.

Route tests (request in → status and body out) survive the inner layers being rearranged. Service
tests, bound to the implementation, are exactly what a refactor breaks and forces you to rewrite.
So the 2,003 uncovered lines in `api/` were both the largest gap and the boundary that has to stay
frozen while we move things around.

## What this changes

| Metric | Before | After |
|---|---|---|
| Line | 63.6% | **74.2%** |
| Branch | 28.2% | **37.4%** |
| `pos/ambrosia/api` | 12.3% | **44.1%** |

128 new tests in 8 files. Suite goes from 745 to **873 tests, 0 failures**.

| Module | Before | After |
|---|---|---|
| `api/Checkout.kt` | 0.0% | 85.2% |
| `api/Authorize.kt` | 0.0% | 84.2% |
| `api/Orders.kt` | 0.0% | 82.9% |
| `utils/AuthUtils.kt` | 66.2% | 81.7% |
| `api/Uploads.kt` | 0.0% | 80.0% |
| `api/Products.kt` | 0.0% | 76.6% |
| `api/Wallet.kt` | 0.0% | 72.9% |
| `api/ProductVariants.kt` | 0.0% | 66.7% |
| `api/Payments.kt` | 0.0% | 66.4% |
| `api/StoreOrders.kt` | 0.0% | 54.5% |

The Kover ratchet in `app/build.gradle.kts` goes from 58/23 to **70/33**, a couple of points below
what was measured. CI already runs `koverVerify`, so the new floor applies on its own.

## Three production bugs this surfaced

All three returned **500** and had never worked. Fixed in `197f35f5`.

1. `GET /orders/{id}/complete` — `CompleteOrder` crossed the HTTP boundary without `@Serializable`.
2. `POST /orders/with-dishes` — same cause, in `OrderWithDishesRequest`.
3. `PUT /orders/{id}/calculate-total` — responded with
   `mapOf("message" to String, "total" to Double)`, a `Map<String, Any>` kotlinx cannot serialize.
   Replaced by a serializable `OrderTotalResponse` that keeps the same JSON shape.

Worth knowing when reviewing: a model missing `@Serializable`, or a heterogeneous `Map`, in a
`respond`/`receive` compiles fine and only fails at runtime. Homogeneous maps do serialize.

## Contracts now pinned

Beyond plain CRUD, these tests freeze behaviour that is easy to lose in a refactor:

- **Wallet scheme separation.** `walletAccessToken` carries `scope=wallet_access` on an
  `HttpOnly`/`SameSite=Strict` cookie; logout revokes against `users.wallet_token`, not just by
  signature. `POST /wallet/invoice` is the declared exception that works with a plain `accessToken`.
- **Login rate limiter.** Blocks after 5 failures with `Retry-After`; a correct login does not get
  through the block, and a successful one resets the counter.
- **Legacy 4-digit PINs** keep working — the 6-digit rule belongs to user creation and editing,
  `/auth/login` must never validate length.
- **`requireAdmin()` reads the `refreshToken` cookie**, not the already-validated JWT principal.
- **Uploads are unauthenticated only before initial setup**, and the client filename is discarded
  (only the extension survives), so there is no traversal.
- **`POST /products/stock` is guarded by `orders_create`**, not `products_update`.
- **The checkout discount check is an inline `requirePermission`**, not a route decorator.
- **Checkout idempotency** — replaying a `paymentHash` returns the original sale (200) instead of
  creating a second one; an unsettled invoice answers 202 `pending`, not a sale.

## Production changes

Kept to the minimum needed to open test seams — no behaviour changes beyond the three fixes above.

- `api/Uploads.kt` — extracted `Route.uploads(uploadService, configService)`; `configureUploads()`
  now calls it. It was the only route module with no seam, so its upload path could not be tested.
- `api/Authorize.kt` — `LoginRateLimiter` is now `internal` with a `resetAll()`. It is
  process-global state keyed by IP, and under `testApplication` every request shares one address,
  so without a reset the tests become order-dependent. Injecting it into `Route.auth(...)` would be
  the better long-term fix.

## Test infrastructure

New `app/src/test/kotlin/pos/ambrosia/utils/RouteTestSupport.kt` — `routeTest` (temp SQLite plus
guaranteed cleanup), `installRoutes`, `grantPermissions`, `jsonBody`, and PBKDF2 credential seeding.
`AdminAuthTestFixture` gains `installWalletAuth`, and `FakeLightningBackend` now records calls and
can simulate failures and unsettled invoices.

Tests mount the `Route.xxx(...)` seam directly with controlled dependencies and skip
`configureXxx()`. `server/docs/TEST_COVERAGE.md` documents the pattern, the available helpers, and
the two global-state traps.

## Still pending

`server/docs/TEST_COVERAGE.md` carries the full ranking. Highlights:

- Some P1 modules are covered but **happy-path only**. Five endpoints are never invoked by any
  test — `POST /store/orders/{id}/refund` is the notable one, since it is the money path.
- **P2**: `AdminNotifications` (2.5%), `InitialSetup` (21%), `Reports` and `Shifts` (0%), `Users`
  (34%), `Roles` (33%). Watch `Handler.kt`: 50.5% line but **5.3% branch**, and it decides which
  status every exception becomes.
- **P4**: `nwc/NostrEvent.createSignedEvent` is at 0% — Schnorr signing on the money path when NWC
  is the active backend.

Note for reviewers: roughly 45 lines in `api/` are path-parameter guards
(`?: respond(400, "Missing…")`) that Ktor's router can never reach, because it does not match the
route when the segment is absent. They are dead code, not pending coverage, and are worth deleting
during the refactor.

## Verification

```bash
cd server
./gradlew test            # 873 tests, 0 failures
./gradlew :app:koverVerify
./gradlew ktlintCheck
```

Each new class also passes on its own (`--tests <Class> --rerun-tasks`), which is where a badly
closed global-state seam would show up.

### DIFF
--- a/server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
+++ b/server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
@@ -1,0 +1,143 @@
+package pos.ambrosia.utest
+
+import io.ktor.client.request.get
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.HttpHeaders
+import io.ktor.http.HttpStatusCode
+import io.ktor.server.response.respond
+import io.ktor.server.routing.get
+import io.ktor.server.routing.routing
+import io.ktor.server.testing.ApplicationTestBuilder
+import pos.ambrosia.utils.AuthCookies
+import pos.ambrosia.utils.authenticateAdmin
+import pos.ambrosia.utils.authorizeAdminPermission
+import pos.ambrosia.utils.authorizePermission
+import pos.ambrosia.utils.getCurrentUser
+import pos.ambrosia.utils.grantPermissions
+import pos.ambrosia.utils.installAdminAuth
+import pos.ambrosia.utils.installNonAdminAuth
+import pos.ambrosia.utils.installRoutes
+import pos.ambrosia.utils.routeTest
+import pos.ambrosia.utils.withAuthCookies
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class AuthUtilsTest {
+    @Test
+    fun `authorizePermission denies a role without the named permission`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            val response = client.get("/guarded") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authorizePermission allows a role holding the named permission`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            grantPermissions("non-admin-test-role", "things_read")
+
+            val response = client.get("/guarded") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.OK, response.status)
+        }
+
+    @Test
+    fun `authorizePermission denies a permission that is not the one required`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            grantPermissions("non-admin-test-role", "things_write")
+
+            val response = client.get("/guarded") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authorizePermission rejects a request with no access token`() =
+        decoratorTest({ installNonAdminAuth() }) {
+            val response = client.get("/guarded")
+
+            assertEquals(HttpStatusCode.Unauthorized, response.status)
+        }
+
+    @Test
+    fun `authenticateAdmin rejects a non-admin role`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            val response = client.get("/admin-only") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authenticateAdmin allows an admin role`() =
+        decoratorTest({ installAdminAuth() }) { auth ->
+            val response = client.get("/admin-only") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.OK, response.status)
+        }
+
+    @Test
+    fun `authenticateAdmin depends on the refreshToken cookie`() =
+        decoratorTest({ installAdminAuth() }) { auth ->
+            val response =
+                client.get("/admin-only") {
+                    headers.append(HttpHeaders.Cookie, "accessToken=${auth.accessToken}")
+                }
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authorizeAdminPermission requires both the permission and the admin flag`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            grantPermissions("non-admin-test-role", "things_read")
+
+            val response = client.get("/admin-guarded") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authorizeAdminPermission allows an admin holding the permission`() =
+        decoratorTest({ installAdminAuth() }) { auth ->
+            grantPermissions("admin-test-role", "things_read")
+
+            val response = client.get("/admin-guarded") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.OK, response.status)
+        }
+
+    @Test
+    fun `getCurrentUser reads the identity claims off the token`() =
+        decoratorTest({ installAdminAuth() }) { auth ->
+            val response = client.get("/whoami") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.OK, response.status)
+            val body = response.bodyAsText()
+            assertTrue(body.contains(auth.userId), "expected the userId claim")
+            assertTrue(body.contains("admin-test-role"), "expected the role claim")
+            assertTrue(body.contains("isAdmin=true"), "expected the isAdmin claim")
+        }
+
+    private fun decoratorTest(
+        installAuth: ApplicationTestBuilder.() -> AuthCookies,
+        block: suspend ApplicationTestBuilder.(AuthCookies) -> Unit,
+    ) = routeTest {
+        val auth = installAuth()
+        installRoutes {
+            routing {
+                authorizePermission("things_read") {
+                    get("/guarded") { call.respond(HttpStatusCode.OK, "ok") }
+                }
+                authenticateAdmin {
+                    get("/admin-only") { call.respond(HttpStatusCode.OK, "ok") }
+                    get("/whoami") { call.respond(HttpStatusCode.OK, call.getCurrentUser().toString()) }
+                }
+                authorizeAdminPermission("things_read") {
+                    get("/admin-guarded") { call.respond(HttpStatusCode.OK, "ok") }
+                }
+            }
+        }
+        block(auth)
+    }
+}
--- a/server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
+++ b/server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
@@ -1,0 +1,309 @@
+package pos.ambrosia.utest
+
+import io.ktor.client.request.delete
+import io.ktor.client.request.get
+import io.ktor.client.request.post
+import io.ktor.client.statement.bodyAsText
+import io.ktor.http.HttpStatusCode
+import io.ktor.server.routing.route
+import io.ktor.server.routing.routing
+import io.ktor.server.testing.ApplicationTestBuilder
+import org.junit.After
+import pos.ambrosia.api.checkout
+import pos.ambrosia.api.storeOrders
+import pos.ambrosia.services.ActiveLightningBackend
+import pos.ambrosia.services.CheckoutService
+import pos.ambrosia.services.RefundService
+import pos.ambrosia.utils.AuthCookies
+import pos.ambrosia.utils.ExposedTestDb
+import pos.ambrosia.utils.FakeLightningBackend
+import pos.ambrosia.utils.grantPermissions
+import pos.ambrosia.utils.installAdminAuth
+import pos.ambrosia.utils.installRoutes
+import pos.ambrosia.utils.jsonBody
+import pos.ambrosia.utils.routeTest
+import pos.ambrosia.utils.withAuthCookies
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+private const val ROLE_NAME = "admin-test-role"
+
+private val allCheckoutPermissions =
+    arrayOf("orders_create", "orders_delete", "orders_refund", "orders_discount")
+
+class CheckoutRoutesTest {
+    private val backend = FakeLightningBackend()
+
+    @After
+    fun tearDown() {
+        ActiveLightningBackend.closeActive()
+    }
+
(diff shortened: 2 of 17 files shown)


server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
```diff
@@ -0,0 +1,143 @@
...
+import pos.ambrosia.utils.withAuthCookies
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+class AuthUtilsTest {
+    @Test
+    fun `authorizePermission denies a role without the named permission`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            val response = client.get("/guarded") { withAuthCookies(auth) }
```
Sharmaz, line 30: Use meaningful variable names instead of `response` (repeated in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
```diff
@@ -0,0 +1,143 @@
...
+
+            assertEquals(HttpStatusCode.Forbidden, response.status)
+        }
+
+    @Test
+    fun `authorizePermission allows a role holding the named permission`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            grantPermissions("non-admin-test-role", "things_read")
+
+            val response = client.get("/guarded") { withAuthCookies(auth) }
```
Sharmaz, line 40: Use meaningful variable names instead of `response`

server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
```diff
@@ -0,0 +1,143 @@
...
+
+            assertEquals(HttpStatusCode.OK, response.status)
+        }
+
+    @Test
+    fun `authorizePermission denies a permission that is not the one required`() =
+        decoratorTest({ installNonAdminAuth() }) { auth ->
+            grantPermissions("non-admin-test-role", "things_write")
+
+            val response = client.get("/guarded") { withAuthCookies(auth) }
```
Sharmaz, line 50: Use meaningful variable names instead of `response`

server/app/src/test/kotlin/pos/ambrosia/utest/AuthUtilsTest.kt
```diff
@@ -0,0 +1,143 @@
...
+            assertEquals(HttpStatusCode.OK, response.status)
+        }
+
+    @Test
+    fun `getCurrentUser reads the identity claims off the token`() =
+        decoratorTest({ installAdminAuth() }) { auth ->
+            val response = client.get("/whoami") { withAuthCookies(auth) }
+
+            assertEquals(HttpStatusCode.OK, response.status)
+            val body = response.bodyAsText()
```
Sharmaz, line 116: Use meaningful variable names instead of `body`, in other documents you have something simillar to `checkoutBody`

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+private const val ROLE_NAME = "admin-test-role"
+
+private val allCheckoutPermissions =
+    arrayOf("orders_create", "orders_delete", "orders_refund", "orders_discount")
+
+class CheckoutRoutesTest {
+    private val backend = FakeLightningBackend()
```
Sharmaz, line 36: Use meaningful variable names instead of `backend`

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+
+    @After
+    fun tearDown() {
+        ActiveLightningBackend.closeActive()
+    }
+
+    @Test
+    fun `a checkout with no items is rejected`() =
+        checkoutTest { auth ->
+            val response =
```
Sharmaz, line 46: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+                client.post("/store/orders/checkout") {
+                    withAuthCookies(auth)
+                    jsonBody(
+                        checkoutBody(items = """[{"productId":"$productId","quantity":1,"priceAtOrder":200}]"""),
+                    )
+                }
+
+            assertEquals(HttpStatusCode.Created, response.status)
+            val body = response.bodyAsText()
```
Sharmaz, line 87: Use meaningful variable names instead of `body`

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+
+            assertEquals(HttpStatusCode.Accepted, response.status)
+            assertTrue(response.bodyAsText().contains("pending"))
+        }
+
+    @Test
+    fun `replaying a payment hash returns the existing checkout`() =
+        checkoutTest { auth ->
+            val productId = ExposedTestDb.seedProduct("Espresso", quantity = 10)
+            val body =
```
Sharmaz, line 118: Use meaningful variable names instead of `body`

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+                checkoutBody(
+                    items = """[{"productId":"$productId","quantity":1,"priceAtOrder":200}]""",
+                    extra = ""","paymentHash":"hash-paid"""",
+                )
+
+            val first =
```
Sharmaz, line 124: Use meaningful variable names instead of `first`

server/app/src/test/kotlin/pos/ambrosia/utest/CheckoutRoutesTest.kt
```diff
@@ -0,0 +1,309 @@
...
+
+    @Test
+    fun `a discounted checkout needs the orders_discount permission`() =
+        routeTest {
+            val auth = installAdminAuth()
+            grantPermissions(ROLE_NAME, "orders_create")
+            mountCheckout()
+            val productId = ExposedTestDb.seedProduct("Espresso", quantity = 10)
+
+            val response =
```
Sharmaz, line 148: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/OrdersRoutesTest.kt
```diff
@@ -0,0 +1,389 @@
...
+private const val ROLE_NAME = "admin-test-role"
+
+class OrdersRoutesTest {
+    @Test
+    fun `listing orders returns the seeded orders`() =
+        ordersTest { auth ->
+            val userId = ExposedTestDb.seedUser("waiter")
+            ExposedTestDb.seedOrder(userId, status = "open")
+
+            val response = client.get("/orders") { withAuthCookies(auth) }
```
Sharmaz, line 35: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/PaymentsRoutesTest.kt
```diff
@@ -0,0 +1,260 @@
...
+
+private const val ROLE_NAME = "admin-test-role"
+
+class PaymentsRoutesTest {
+    @Test
+    fun `listing payments returns the seeded payments`() =
+        paymentsTest { auth ->
+            ExposedTestDb.seedPayment(amount = 33.0)
+
+            val response = client.get("/payments") { withAuthCookies(auth) }
```
Sharmaz, line 35: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/ProductsRoutesTest.kt
```diff
@@ -0,0 +1,288 @@
...
+
+private const val ROLE_NAME = "admin-test-role"
+
+class ProductsRoutesTest {
+    @Test
+    fun `listing products returns the seeded products`() =
+        productsTest { auth ->
+            ExposedTestDb.seedProduct("Espresso")
+
+            val response = client.get("/products") { withAuthCookies(auth) }
```
Sharmaz, line 36: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/UploadsRoutesTest.kt
```diff
@@ -0,0 +1,130 @@
...
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class UploadsRoutesTest {
+    private lateinit var uploadRoot: Path
+
+    @Test
+    fun `upload without a token is allowed before the config row exists`() =
+        uploadTest {
+            val response = client.post("/uploads") { setBody(filePart("logo.png")) }
```
Sharmaz, line 38: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/WalletRoutesTest.kt
```diff
@@ -0,0 +1,435 @@
...
+                }
+
+            assertEquals(HttpStatusCode.OK, bumpResponse.status)
+            assertEquals(HttpStatusCode.OK, exportResponse.status)
+        }
+
+    @Test
+    fun `an undecodable invoice is a bad request`() =
+        walletTest { auth ->
+            val response =
```
Sharmaz, line 222: Use meaningful variable names instead of `response` (in all the document)

server/app/src/test/kotlin/pos/ambrosia/utest/WalletRoutesTest.kt
```diff
@@ -0,0 +1,435 @@
...
+                        isPaid = true,
+                        receivedSat = 1500,
+                        fees = 0,
+                        createdAt = 0,
+                    ),
+                )
+
+            val response = client.get("/wallet/payments/incoming") { withAuthCookies(auth) }
+
+            val body = response.bodyAsText()
```
Sharmaz, line 346: Use meaningful variable names instead of `body`

server/app/src/test/kotlin/pos/ambrosia/utils/RouteTestSupport.kt
```diff
@@ -0,0 +1,110 @@
...
+            ?.toString()
+            ?: ExposedTestDb.seedPermission(name)
+    }
+
+fun setRolePassword(
+    roleId: String,
+    password: String,
+    secret: String = TEST_SECRET,
+) {
+    val hash = SecurePinProcessor.hashPinForStorage(password.toCharArray(), roleId, testEnvironment(secret))
```
Sharmaz, line 85: Use meaningful variable names instead of `hash` (in all the document)

Sharmaz (changes requested):

PR #738, opened 2026-08-28
Feature/data export import

Sharmaz:
## Changes:

- Add encrypted business-data export (`POST /backup/export`, server) and an Export Data card in Settings that triggers it and downloads the resulting file, gated by the wallet password.
- Add business-data import (`POST /backup/import`, server) with a pending-import/secret swap applied on the next restart, an Import Data card in Settings with a destructive-action confirmation, and a restore-from-backup option in onboarding for restoring onto a fresh install (`POST /initial-setup/restore`).
- Restart the Electron backend automatically after an import, and show a blocking restart-required modal instead of an easy-to-miss toast when running outside Electron.
- Clear device sessions, wallet tokens, and push subscriptions on server startup after a pending import is applied, so stale sessions from the previous business don't linger.
- Fix the imported keystore being left undecryptable after a restart, by deleting it on secret swap so it regenerates with the new secret.
- Require a separate backup password from the role/wallet password when importing, since a backup's own password can differ from the current business's — split across the import endpoint, the client service, and both entry points' UI.
- Report real export/import progress (byte-based) instead of an indeterminate spinner, across the export route, both service functions, the Export/Import Data cards, and the onboarding restore step.
- Extract a shared `BackupPasswordAndFileFields` component for the backup password and file inputs, used by both the Import Data card and the onboarding restore step.
- Polish: move the onboarding restore-link above the divider with a proper hover affordance, restyle the file-choice trigger as a primary-action button, align button styles with the rest of the app, remove locale-string interpolation in favor of composing text in JSX, and reword the backup-password prompts (and add an export-time notice) so they reference the wallet password instead of unexplained "encryption".

**Screenshots**:

Onboarding

Screenshot 2026-08-28 at 1 56 16 p m
Screenshot 2026-08-28 at 1 56 30 p m
Screenshot 2026-08-28 at 1 56 59 p m
Screenshot 2026-08-28 at 1 57 08 p m

Settings

Screenshot 2026-08-28 at 1 53 35 p m
Screenshot 2026-08-28 at 1 54 21 p m
Screenshot 2026-08-28 at 1 54 57 p m
Screenshot 2026-08-28 at 1 55 08 p m
Screenshot 2026-08-28 at 1 55 18 p m


server/app/src/main/kotlin/pos/ambrosia/services/BackupService.kt
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    private fun writeFileEntry(
+        zip: ZipOutputStream,
+        source: Path,
+        entryName: String,
+    ) {
+        zip.putNextEntry(ZipEntry(entryName))
+        Files.newInputStream(source).use { it.copyTo(zip) }
+        zip.closeEntry()
```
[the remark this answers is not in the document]
Sharmaz: **False positive.** `writeFileEntry` doesn't call `ZipOutputStream.write(...)` directly — it copies via `Files.newInputStream(source).use { it.copyTo(zip) }`. Kotlin's `InputStream.copyTo` reads the source into a buffer in a loop and calls `write(buffer, offset, length)` on the destination stream internally, so **the entry's content is written** — CodeQL just isn't tracing the call through the stdlib helper.

Confirmed with a **real end-to-end test**, not just static reasoning: exporting a business and re-importing that same file restores actual data (users, products, orders) with no corruption or truncation. An empty entry here would mean an empty `ambrosia.db` in every export, which contradicts what we've verified working.

**Not changing the code** to satisfy the rule — swapping `copyTo` for a manual `read`/`write` loop would make this less idiomatic for no functional gain. Dismissing as a false positive.

client/src/services/backupService.js
```diff
@@ -0,0 +1,109 @@
...
+  return error;
+}
+
+function extractFilename(contentDisposition, fallbackFilename) {
+  const match = contentDisposition?.match(/filename="?([^"]+)"?/);
+  return match?.[1] ?? fallbackFilename;
+}
+
+async function readResponseWithProgress(response, totalBytes, onProgress) {
+  if (!response.body || !totalBytes || !onProgress) {
```
Vidarte-Alberto: Please use more descriptive names, response variable is not descriptive

server/app/src/main/kotlin/pos/ambrosia/Api.kt
```diff
@@ -65,6 +67,9 @@ public val logger = LoggerFactory.getLogger("Server")
 class Api {
     fun Application.module() {
         AppConfig.loadConfig() // Load the configuration
```
Vidarte-Alberto: please remove unnecessary comments

Vidarte-Alberto (changes requested):
Please do fixes

Vidarte-Alberto (approved):
LGTM!

ca-ruz (approved):
LGTM!

Built and tested locally, works great!

Image

Image

Image

Image

PR #744, opened 2026-09-01
Feature/backup progress direct websocket

Sharmaz:
## Changes:

- Add a scoped WebSocket channel (`/ws/backup-progress`), authenticated with a short-lived, single-use token, so the browser can receive real-time backup progress directly from the server.
- Report real byte-level progress (not just transfer bytes) from `BackupService` during export, import, and restore, and wire it into `/backup/export`, `/backup/import`, and `/initial-setup/restore`.
- Wire the new progress channel into the client — Settings' Export/Import cards and the onboarding restore step — replacing the previous byte-tracking-only indicators.
- Add a dedicated, unauthenticated `/initial-setup/progress-token` endpoint so onboarding restore also gets real-time progress, since the existing `/backup/progress-token` requires a wallet session that doesn't exist yet at that point in the flow.
- Add staleness protection for pending imports: a staged-but-never-applied import is discarded automatically after 24 hours instead of being silently applied by an unrelated server restart, and a new import is rejected while a fresh one is already pending.
- Show the server's specific error message instead of a generic one when an import/restore is rejected for a pending-import conflict, in both the Settings and onboarding flows.
- Fix the Import Data card in Settings getting stuck showing a stale progress percentage after a successful import.

**Screenshots**:

Screenshot 2026-08-31 at 6 03 28 p m

Screenshot 2026-08-31 at 5 59 54 p m

Screenshot 2026-09-01 at 3 46 47 p m

Screenshot 2026-09-01 at 3 49 18 p m


ca-ruz (approved):
ACK.

Image

Image

server/app/src/main/kotlin/pos/ambrosia/services/BackupService.kt
```diff
@@ -199,11 +242,27 @@ class BackupService(
...
+
     private fun extractStagedBackup(
         zip: ZipInputStream,
         stagingTempRoot: Path,
+        onProgress: (phase: String, bytesProcessed: Long, totalBytes: Long?) -> Unit,
     ): BackupManifest {
         var importedManifest: BackupManifest? = null
+        var bytesExtractedSoFar = 0L
         zip.use {
             var entry = zip.nextEntry
```
Vidarte-Alberto: Please use more decriptive variables names. Please change entry variable name

Vidarte-Alberto (changes requested):
Please do fixes

Vidarte-Alberto (approved):
LGTM!!

PR #745, opened 2026-09-03
Fix/import apply restart token

Sharmaz:
## Changes:

- Add a signed confirmation token (`TokenService.generateBackupConfirmationToken`/`isBackupConfirmationTokenValid`) and require it in `BackupService.applyPendingImport` before a staged import/restore is ever applied, so a server restart no longer applies data that nobody explicitly confirmed.
- Expose the running `EmbeddedServer` and Docker mode in `Ambrosia.kt`/`DockerMode.kt` so a confirmed import can trigger a graceful, automatic restart when running in Docker.
- Add `POST /backup/confirm-pending-import` and `POST /initial-setup/confirm-pending-restore` endpoints that write the confirmation token and, in Docker, schedule the restart.
- Add `confirmPendingImport`/`confirmPendingRestore` to the client's `backupService`/`initialSetupService`.
- Call the new confirm functions before triggering a backend restart in both the Settings Import Data card and the onboarding restore step.


ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):
LGTM!!

PR #748, opened 2026-09-09
Feat/native(chore): restart controls

### LINKED ISSUE #741: Add owner-only recovery controls for Lightning, Ambrosia, and device restart/shutdown
## Summary

Add owner/admin-only recovery controls for local deployments:

- refresh/restart phoenixd Lightning backend;
- restart Ambrosia backend/service;
- reboot the device/OPI;
- gracefully shut down the device/OPI.

These actions should help field operators recover from temporary service or Lightning state problems without SSH, while keeping the controls gated and auditable.

## Context

Recent investigation suggests that repeated outgoing payment failures may have been resolved by a restart/reconnection/state refresh rather than by a phoenixd binary upgrade.

Ambrosia already has related Electron/service-management code:

- `electron/main.js` handles `services:restart`.
- `electron/services/ServiceManager.js` has `restartService(serviceName)`.
- `electron/main.js` has `phoenixd:set-auto-liquidity`, which can call `serviceManager.restartService('phoenixd')`.
- Settings UI already has auto-liquidity restart copy, but that is not a general payment recovery action.

The new recovery controls should be separate from auto-liquidity settings because changing liquidity configuration is not the same as recovering from a stuck/temporary outgoing payment issue.

## Proposed Work

Add a protected recovery section with these actions:

- `Refresh Lightning`: restart phoenixd only, when Ambrosia manages phoenixd locally.
- `Restart Ambrosia`: restart only the Ambrosia backend/service.
- `Restart Device`: reboot the OPI/device.
- `Shut Down Device`: gracefully shut down the OPI/device.

Requirements:

- owner/admin only;
- not shown to cashier role by default;
- confirmation modal for each action;
- clear copy explaining impact;
- disable actions while one is already in progress;
- record audit event with actor/timestamp/action/result;
- platform guard so actions are available only where supported;
- hide or disable Lightning restart when using NWC or an external backend Ambrosia cannot restart;
- after phoenixd restart, poll phoenixd health/getinfo/getbalance/listchannels;
- after Ambrosia restart, show reconnect/progress state if the UI remains open.

## UX Guidance

Avoid wording such as:

> Fix phoenixd issue

Prefer operational wording:

> Refresh Lightning backend

For Lightning refresh, explain:

> This restarts the local phoenixd service. Pending Lightning payments may need time to settle. Use this when routing or temporary peer errors repeat.

For device reboot/shutdown, clearly state that the POS may be unavailable until the device is back online.

## Acceptance Criteria

- Owner/admin can restart managed phoenixd from Ambrosia.
- Owner/admin can restart Ambrosia where supported.
- Owner/admin can reboot and gracefully shut down the OPI/device where supported.
- Cashier/non-admin cannot access these actions.
- Unsupported environments show a clear unavailable state.
- NWC/external Lightning backend does not show a misleading phoenixd restart action.
- Each action requires confirmation and prevents duplicate execution.
- Success/failure states are visible to the user.
- Audit logs capture actor, timestamp, action, and result.
- Tests cover permissions, confirmations, success, failure, unsupported environment, NWC/external backend, and duplicate action prevention.

## Security Notes

These actions must not become unauthenticated local endpoints. They should require authenticated owner/admin permissions and should not be available from public/customer views.

Vidarte-Alberto:
## Problem

Ambrosia had no way to restart the server or phoenixd from the app itself. If either process needed to pick up a config change, recover from a stuck state, or apply a pending settings change, the only options were a manual terminal/systemctl command (native installs), physically rebooting the device (hardware), or restarting containers by hand (Docker) — with zero in-app control. Electron already had the plumbing for this internally (`services:restart` IPC), but it was never exposed as an actual button, and no other deployment mode had anything at all.

## Fix

Added "Restart server" / "Restart phoenixd" controls to Settings, working across every deployment mode — native (systemd), the Raspberry Pi/OrangePi hardware image, Docker, and Electron — by reusing each mode's own existing process supervisor (systemd `Restart=always`, Docker `restart: unless-stopped`, Electron's `ServiceManager`) instead of introducing new privileges like passwordless sudo or polkit rules. phoenixd is signalled directly via a pidfile since it runs as the same OS user as the server on native/hardware installs; on Docker it correctly has no way to be reached (separate container, separate PID namespace) so only the server button is shown there.

## What Changed

- Added `canRestartSelf()`/`isSystemdManaged()`/`isLocalPhoenixd()` server helpers to detect when a self-restart is safe.
- Generalized the existing Docker-only restart (`scheduleDockerRestart`) into `scheduleProcessRestart`, now also used for systemd-managed native/hardware installs.
- Forced the process to `exitProcess` after stopping the embedded server — a leaked non-daemon thread from the Web Push async HTTP client was silently keeping the JVM alive after `stopSuspend()`, so the process never actually restarted.
- Added `PhoenixdProcessService`, which signals phoenixd through a pidfile.
- Added admin-only `/system/restart-capabilities`, `/system/restart-server`, `/system/restart-phoenixd` endpoints.
- Added `run-server.sh` / `run-phoenixd.sh` self-restart wrappers with signal forwarding, so `systemctl stop` still stops the process cleanly instead of fighting the restart loop.
- Wired the phoenixd wrapper and an `AMBROSIA_SERVICE_MANAGED` flag into `install.sh` and the hardware image's systemd units.
- Added an `install.sh --local` flag that builds and installs from the current checkout instead of downloading a GitHub release; updated `uninstall.sh` to remove the new wrapper too.
- Added a "System" Settings card with a confirmation modal for each restart action, styled to match the Close Channel danger buttons.
- Fixed two unrelated pre-existing bugs found while testing this end-to-end: a malformed fallback wordlist URL in `SeedGenerator` (a stray triple-quoted string embedded newlines into the URL) that broke server boot on any fresh install without the local wordlist file, and the Docker client not binding to `0.0.0.0`, which made its own middleware unable to reach itself for the onboarding check.

## How to Test

1. **Native**: ` ./scripts/install.sh --local --yes`, log in as admin, go to Settings → System, click "Restart server" and confirm — the service should actually come back. Same for "Restart phoenixd".
2. **Hardware**: build and flash the OrangePi/RPi image, boot the device, repeat the same two checks.
3. **Docker**: `docker compose up --build -d`, confirm only "Restart server" is shown (not phoenixd), and that clicking it restarts the `ambrosia` container.
4. **Electron**: confirm both buttons work via the existing `services:restart` IPC channel.
5. Confirm neither button appears on a deployment with no supervisor (e.g. a `--no-service` install).

## Verification

- Server: `./gradlew test ktlintCheck` passed.
- Client: full Jest suite passed (306 suites, 2905 tests), ESLint clean.
- All new/modified shell scripts pass `bash -n`.
- Manually verified end-to-end on native, hardware (real device flash + boot), Docker, and Electron.

## Screenshots

image

### DIFF
--- a/client/src/components/pages/Store/Settings/System/__tests__/SystemCard.test.jsx
+++ b/client/src/components/pages/Store/Settings/System/__tests__/SystemCard.test.jsx
@@ -1,0 +1,165 @@
+import { render, screen, fireEvent, act } from "@testing-library/react";
+
+import * as useSystemRestartHook from "@/hooks/useSystemRestart";
+
+import { SystemCard } from "../SystemCard";
+
+jest.mock("@/hooks/useSystemRestart");
+
+jest.mock("@heroui/react", () => ({
+  addToast: jest.fn(),
+  Button: ({ onPress, children, isLoading, ...props }) => (
+    <button type="button" disabled={isLoading} onClick={onPress} {...props}>{children}</button>
+  ),
+  Card: ({ children }) => <div>{children}</div>,
+  CardHeader: ({ children }) => <div>{children}</div>,
+  CardBody: ({ children }) => <div>{children}</div>,
+}));
+
+jest.mock("../RestartConfirmModal", () => ({
+  RestartConfirmModal: ({ isOpen, title, description, onConfirm, onCancel }) => (
+    isOpen ? (
+      <div data-testid="restart-confirm-modal">
+        <span>{title}</span>
+        <span>{description}</span>
+        <button type="button" onClick={onConfirm}>confirm</button>
+        <button type="button" onClick={onCancel}>cancel</button>
+      </div>
+    ) : null
+  ),
+}));
+
+function mockUseSystemRestart(overrides = {}) {
+  const defaults = {
+    serverRestartSupported: false,
+    phoenixdRestartSupported: false,
+    restartingTarget: null,
+    loadRestartCapabilities: jest.fn(),
+    restartServer: jest.fn().mockResolvedValue(true),
+    restartPhoenixd: jest.fn().mockResolvedValue(true),
+  };
+  const state = { ...defaults, ...overrides };
+  jest.spyOn(useSystemRestartHook, "useSystemRestart").mockReturnValue(state);
+  return state;
+}
+
+describe("SystemCard", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe("Rendering", () => {
+    it("renders the title and description", () => {
+      mockUseSystemRestart();
+      render(<SystemCard />);
+
+      expect(screen.getByText("cardSystem.title")).toBeInTheDocument();
+      expect(screen.getByText("cardSystem.description")).toBeInTheDocument();
+    });
+
+    it("shows the unsupported notice when neither target can be restarted", () => {
+      mockUseSystemRestart();
+      render(<SystemCard />);
+
+      expect(screen.getByText("cardSystem.unsupportedNotice")).toBeInTheDocument();
+      expect(screen.queryByText("cardSystem.restartServerButton")).not.toBeInTheDocument();
+      expect(screen.queryByText("cardSystem.restartPhoenixdButton")).not.toBeInTheDocument();
+    });
+
+    it("shows only the server button when only server restart is supported", () => {
+      mockUseSystemRestart({ serverRestartSupported: true });
+      render(<SystemCard />);
+
+      expect(screen.getByText("cardSystem.restartServerButton")).toBeInTheDocument();
+      expect(screen.queryByText("cardSystem.restartPhoenixdButton")).not.toBeInTheDocument();
+    });
+
+    it("shows both buttons when both targets are supported", () => {
+      mockUseSystemRestart({ serverRestartSupported: true, phoenixdRestartSupported: true });
+      render(<SystemCard />);
+
+      expect(screen.getByText("cardSystem.restartServerButton")).toBeInTheDocument();
+      expect(screen.getByText("cardSystem.restartPhoenixdButton")).toBeInTheDocument();
+    });
+
+    it("loads restart capabilities on mount", () => {
+      const state = mockUseSystemRestart();
+      render(<SystemCard />);
+
+      expect(state.loadRestartCapabilities).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe("Restarting the server", () => {
+    it("opens the confirmation modal when the restart server button is pressed", () => {
+      mockUseSystemRestart({ serverRestartSupported: true });
+      render(<SystemCard />);
+
+      fireEvent.click(screen.getByText("cardSystem.restartServerButton"));
+
+      expect(screen.getByTestId("restart-confirm-modal")).toBeInTheDocument();
+      expect(screen.getByText("cardSystem.confirmServerTitle")).toBeInTheDocument();
+    });
+
+    it("calls restartServer and shows a success toast on confirm", async () => {
+      const state = mockUseSystemRestart({ serverRestartSupported: true });
+      const { addToast } = require("@heroui/react");
+      render(<SystemCard />);
+
+      fireEvent.click(screen.getByText("cardSystem.restartServerButton"));
+      await act(async () => {
+        fireEvent.click(screen.getByText("confirm"));
+      });
+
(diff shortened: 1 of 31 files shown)


client/src/components/pages/Store/Settings/System/__tests__/SystemCard.test.jsx
```diff
@@ -0,0 +1,165 @@
...
+    serverRestartSupported: false,
+    phoenixdRestartSupported: false,
+    restartingTarget: null,
+    loadRestartCapabilities: jest.fn(),
+    restartServer: jest.fn().mockResolvedValue(true),
+    restartPhoenixd: jest.fn().mockResolvedValue(true),
+  };
+  const state = { ...defaults, ...overrides };
+  jest.spyOn(useSystemRestartHook, "useSystemRestart").mockReturnValue(state);
+  return state;
```
Sharmaz: Avoid generic variables as `state` in the all file.

client/src/hooks/__tests__/useSystemRestart.test.js
```diff
@@ -0,0 +1,100 @@
...
+    it("applies the capabilities returned by the server", async () => {
+      getRestartCapabilities.mockResolvedValue({
+        serverRestartSupported: true,
+        phoenixdRestartSupported: false,
+      });
+      const { result } = renderHook(() => useSystemRestart());
+
+      await act(async () => {
+        await result.current.loadRestartCapabilities();
+      });
```
Sharmaz: Avoid generic variables as `result` in the whole document.

client/src/hooks/__tests__/useSystemRestart.test.js
```diff
@@ -0,0 +1,100 @@
...
+  });
+
+  describe("restartServer", () => {
+    it("calls the server restart endpoint and returns true on success", async () => {
+      restartServer.mockResolvedValue({ message: "Server restarting" });
+      const { result } = renderHook(() => useSystemRestart());
+
+      let returnValue;
+      await act(async () => {
+        returnValue = await result.current.restartServer();
```
Sharmaz: Be more specific with the variable names `returnValue` is meaningless

server/app/src/test/kotlin/pos/ambrosia/utest/PhoenixdProcessServiceTest.kt
```diff
@@ -0,0 +1,45 @@
...
+        assertEquals(PhoenixdRestartResult.NoPidFile, service.requestRestart())
+        assertFalse(service.isRunning())
+    }
+
+    @Test
+    fun `requestRestart returns NotRunning when the pidfile references a dead process`() {
+        val deadProcess = ProcessBuilder("true").start()
+        deadProcess.waitFor()
+        val pidFile = newPidFile().apply { writeText(deadProcess.pid().toString()) }
+        val service = PhoenixdProcessService(pidFile = pidFile)
```
Sharmaz: Avoid generic variables as `service` in the whole document.

server/app/src/test/kotlin/pos/ambrosia/utest/ServiceSupervisionTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        testApplication {
+            application { routing { get("/probe") { call.respondText(call.isLocalPhoenixd().toString()) } } }
+
+            assertEquals("true", client.get("/probe").bodyAsText())
+        }
+
+    @Test
+    fun `isLocalPhoenixd is false when phoenixd-remote is enabled`() =
+        testApplication {
+            environment { config = MapApplicationConfig("phoenixd-remote" to "true") }
```
Sharmaz, line 28: Avoid generic variables as `config` in the whole document.

Sharmaz (changes requested):

server/app/src/test/kotlin/pos/ambrosia/utest/ServiceSupervisionTest.kt
Vidarte-Alberto: Config here isn't a local variable we're naming, it's the config property on Ktor's ApplicationEngineEnvironmentBuilder, set via the environment { config = ... } test DSL (ktor-server-test-host).

Sharmaz (approved):
LGTM!

ca-ruz (approved):
ACK.

Image

Phoenixd Restart:

Image

Image

Image


Server Restart:

Image

Image

Image

PR #751, opened 2026-09-11
Feat(server): freelance reports api

### LINKED ISSUE #656: Freelancer Reports service + API
`/freelancer/reports` needs an aggregated, exportable view by project/task for a given period — nothing computes this today.

`FreelancerReportService` + `configureFreelancerReports()`, `GET /reports?client_id&from&to`, returning hours + amounts grouped by project/task.

✅ **Product decision folded in (2026-07-15, multi-currency UX)**: when a query spans clients billed in different currencies, the API returns raw totals broken down per currency (no fiat-to-fiat conversion — none exists in the codebase). The BTC-equivalent aggregate total is computed client-side by the Reports page (see its note), not here — keep this endpoint's response currency-neutral per line.

None discussed beyond noting this shares logic conceptually with the Invoices API's invoice-generation aggregation — share a query/helper between the two rather than duplicating the "billable time_entries in a date range" logic.

**Additional context**
- Area: Backend. Size: M.
- Depends on: "Time Entries service + API + rate resolution". Blocks: "Reports page".
- Files: `server/.../services/FreelancerReportService.kt`, `server/.../api/FreelancerReports.kt`.
- Acceptance criteria: report totals for a period match what the Invoices API's aggregation would roll up for the same period/client — cross-check once both exist.

Vidarte-Alberto:
## Summary

Implements the freelance billing reports API for issue #<fill-in-issue-number>. Adds an endpoint that aggregates billable hours and amounts by project/task for a date range, so `/freelancer/reports` on the frontend has something to render.

## Changes

- Adds `BillableTimeAggregationService`, a standalone service that resolves billable time entries in a date range: rate falls back from project → client, rate/amount are zeroed for non-billable entries, and amounts are rounded with `BigDecimal` (HALF_UP) to avoid rounding drift. Kept as its own service (not private to the report) so the future Invoices API can call the same source of truth instead of duplicating this logic.
- Adds `FreelanceReportService`, which groups those resolved lines into `currency → project → task`. Currency groups are never converted or combined — when a query spans clients billed in different currencies, each currency gets its own totals, matching the product decision that fiat-to-fiat conversion belongs client-side.
- Adds `GET /freelance/billing-reports?from&to&client_id` (`client_id` optional), registered via `configureFreelanceReports()` and guarded by the existing `freelance_reports_read` permission (already seeded in `V45__add_freelance_permissions.sql` — no new migration needed).
- Adds `FreelanceReportResponse` / `FreelanceReportCurrencyGroup` / `FreelanceReportProjectGroup` / `FreelanceReportTaskGroup` response models.
- Non-billable time entries are included in the report (with `amountCents = 0`) so freelancers can see total hours worked, not only billable ones.

## Out of scope

- The frontend Reports page (client-side BTC-equivalent aggregation and CSV export).
- The Invoices API and its invoice-generation aggregation — `BillableTimeAggregationService` is built to be reused by it once it exists, but cross-checking totals against it isn't possible yet.

## Verification

- `./gradlew test --tests "*BillableTimeAggregation*" --tests "*FreelanceReport*"`
- `./gradlew ktlintCheck test`

### DIFF
--- a/server/app/src/test/kotlin/pos/ambrosia/utest/FreelanceReportsRouteTest.kt
+++ b/server/app/src/test/kotlin/pos/ambrosia/utest/FreelanceReportsRouteTest.kt
@@ -37,7 +37,8 @@ class FreelanceReportsRouteTest {
     @Test
     fun `route requires authentication and the matching permission`() =
         testApplication {
-            val auth = installNonAdminAuth("freelance-reports-no-permission", "freelance-reports-no-permission-user")
+            val authWithoutPermission =
+                installNonAdminAuth("freelance-reports-no-permission", "freelance-reports-no-permission-user")
             application {
                 install(ContentNegotiation) { json() }
                 handler()
@@ -52,7 +53,7 @@ class FreelanceReportsRouteTest {
                 HttpStatusCode.Forbidden,
                 client
                     .get("/freelance/billing-reports?from=2026-08-17&to=2026-08-23") {
-                        withAuthCookies(auth)
+                        withAuthCookies(authWithoutPermission)
                     }.status,
             )
         }
@@ -60,7 +61,7 @@ class FreelanceReportsRouteTest {
     @Test
     fun `missing from or to returns bad request`() =
         testApplication {
-            val auth = installNonAdminAuth("freelance-reports-invalid", "freelance-reports-invalid-user")
+            val authWithPermission = installNonAdminAuth("freelance-reports-invalid", "freelance-reports-invalid-user")
             grantPermission("freelance-reports-invalid", "freelance_reports_read")
             application {
                 install(ContentNegotiation) { json() }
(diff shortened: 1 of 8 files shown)


server/app/src/test/kotlin/pos/ambrosia/utest/FreelanceReportsRouteTest.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    @After
+    fun tearDown() {
+        ExposedTestDb.cleanup(databaseFile)
+    }
+
+    @Test
+    fun `route requires authentication and the matching permission`() =
+        testApplication {
+            val auth = installNonAdminAuth("freelance-reports-no-permission", "freelance-reports-no-permission-user")
```
Sharmaz: Avoid generic names as `auth`

server/app/src/test/kotlin/pos/ambrosia/utest/FreelanceReportsRouteTest.kt
```diff
@@ -0,0 +1,110 @@
...
+            val projectId = ExposedTestDb.seedProject(clientId, name = "Project Alpha")
+            val taskId = ExposedTestDb.seedTask("Development")
+            ExposedTestDb.seedTimeEntry(projectId, taskId, entryDate = "2026-08-19", durationMinutes = 60)
+            application {
+                install(ContentNegotiation) { json() }
+                handler()
+                configureFreelanceReports()
+            }
+
+            val response =
```
Sharmaz: Avoid generic names as `response`

Sharmaz (changes requested):

ca-ruz (approved):
LGTM!

Sharmaz (approved):
LGTM!!

PR #752, opened 2026-09-11
Fix/electron backup import reload

Sharmaz:
## Changes:

- Move the Electron-vs-manual restart decision out of each `RestartRequiredModal` caller and into the component itself, removing duplicated `isElectron` branching across 3 call sites (`PhoenixdRemoteCardUnlocked`, `ImportData`, `RestoreFromBackup`).
- Relaunch the app in Electron (instead of only restarting the backend) after importing a backup in Settings or restoring one during onboarding, so the UI actually reflects the freshly-imported data instead of silently leaving the user on stale screen.

### Screenshots

Screenshot 2026-09-11 at 10 43 01 a m


ca-ruz (approved):
LGTM!

Vidarte-Alberto (approved):
LGTM!!

PR #755, opened 2026-09-12
Chore/electron testability

Sharmaz:
## Changes:

- Add Vitest as the electron test runner, with a require.cache-based mock for the `electron` module (it resolves to a binary path string outside a real Electron process, so no standard mocking approach reaches it) and a first test covering `utils/logger.js`.
- Cover `utils/resourcePaths.js`, `utils/portAllocator.js`, `utils/healthCheck.js`, `preload.entry.js`, and `splash-preload.entry.js`, extending the electron mock to also cover `contextBridge`/`ipcRenderer`.
- Cover the services layer (`ConfigurationBootstrap`, `PhoenixdService`, `BackendService`, `NextJsService`, `ServiceManager`, `AutoUpdater`), adding shared spawn/tree-kill/platform mocks reused across the new service tests and refactored into the existing `resourcePaths` tests.
- Cover `main.js`'s IPC handlers, admin-activity notification, and single-instance lock without modifying the file itself.
- Add a GitHub Actions workflow to run the electron test suite on every pull request.

### DIFF
--- a/electron/eslint.config.mjs
+++ b/electron/eslint.config.mjs
@@ -149,6 +149,23 @@ const eslintConfig = [
       'import/prefer-default-export': 'off',
     },
   },
+  {
+    files: ['**/__tests__/**/*.test.js'],
+    languageOptions: {
+      globals: {
+        // vitest globals
+        describe: 'readonly',
+        it: 'readonly',
+        test: 'readonly',
+        expect: 'readonly',
+        vi: 'readonly',
+        beforeAll: 'readonly',
+        afterAll: 'readonly',
+        beforeEach: 'readonly',
+        afterEach: 'readonly',
+      },
+    },
+  },
 ];
 
 export default eslintConfig;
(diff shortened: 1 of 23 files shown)


electron/eslint.config.mjs
```diff
@@ -149,6 +149,23 @@ const eslintConfig = [
       'import/prefer-default-export': 'off',
     },
   },
+  {
+    files: ['**/__tests__/**/*.test.js', 'test-utils/**/*.js'],
+    languageOptions: {
+      globals: {
+        // vitest globals
```
Vidarte-Alberto, line 156: Please remove unnecessary comments.

Vidarte-Alberto (changes requested):
Please do fix

electron/eslint.config.mjs
Sharmaz: At config files, comments are allowed, because there are not naming to identify wich are vitest stuff

Vidarte-Alberto (approved):
LGTM!!!

ca-ruz (approved):
LGTM!

